PR #9, opened 2026-05-14
fix: remove /provider alias, use /model as canonical command

zhujian7:
## Summary
- Remove `/provider` as an alias for `/model` slash command
- Update all "Use /provider to connect" messages to "/model"
- `/model` is now the single canonical command for managing provider connections

## Related issue(s)
N/A

---
🤖 Generated with Claude Code

### DIFF
--- a/cmd/gen/main.go
+++ b/cmd/gen/main.go
@@ -175,8 +175,7 @@ Keybindings:
   Ctrl+C       Clear input / Quit
 
 Slash Commands:
-  /provider    Select and connect to a provider
-  /model       Select a model
+  /model       Select model and manage provider connections
   /clear       Clear chat history
   /help        Show help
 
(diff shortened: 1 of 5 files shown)


yanmxa:
image

yanmxa:
Hi @zhujian7, thanks for the fix! 🙏

Quick context: since provider config is tightly coupled to the model (each model implies its provider + connection setup), we previously deprecated `/provider` and consolidated everything under `/model` to keep the slash command surface small and the mental model consistent — pick a model, provider follows.

That said, the message at `internal/app/input/slash_command.go:629` (`"Use /provider to connect."`) is a real bug — it points at a command that no longer exists. The preferred fix is updating that prompt (and any other stale `/provider` references) to `/model`, rather than re-adding the alias.

Would you be up for repurposing the PR that way? Thanks again! 🙌

zhujian7:
@yanmxa PTAL

PR #11, opened 2026-05-15
feat: add native DeepSeek provider

zhujian7:
## Summary
- Add native DeepSeek API provider (api.deepseek.com) using OpenAI-compatible SDK
- Support `deepseek-v4-flash` and `deepseek-v4-pro` models with reasoning_effort extraction
- Include model catalog with token limits and USD pricing, with cost tracking wired into model.go

## Test plan
- [x] Build passes
- [x] All new unit tests pass (model list error propagation, stream request, cost estimation, thinking support)
- [x] Existing provider tests unaffected
- [x] Manual end-to-end test with valid DEEPSEEK_API_KEY

---
🤖 Generated with Claude Code


yanmxa (commented):
Six inline comments on the items I think need addressing before merge. Full review context in the PR conversation.

internal/llm/deepseek/catalog.go
```diff
@@ -0,0 +1,91 @@
...
+
+type modelCatalogEntry struct {
+	info    llm.ModelInfo
+	pricing pricing
+}
+
+var catalog = []modelCatalogEntry{
+	{
+		info: llm.ModelInfo{
+			ID:               "deepseek-v4-flash",
```
yanmxa, line 20: PR description says V3/R1, but the catalog ships `deepseek-v4-flash` / `deepseek-v4-pro` as primary entries with 1M input / 384K output limits and a specific deprecation date for the legacy names (line 38). These IDs aren't in DeepSeek's published lineup — please cite the source for the V4 family + pricing, or drop the V4 entries and keep only `deepseek-chat` / `deepseek-reasoner`.

internal/llm/deepseek/catalog.go
```diff
@@ -0,0 +1,91 @@
...
+func CatalogModel(modelID string) (llm.ModelInfo, bool) {
+	for _, entry := range catalog {
+		if entry.info.ID == modelID {
+			return entry.info, true
+		}
+	}
+	return llm.ModelInfo{}, false
+}
+
+func EstimateCost(modelID string, usage llm.Usage) (llm.Money, bool) {
```
yanmxa, line 57: `EstimateCost` is dead code — `internal/app/model.go` only switches on `llm.MinMax` for cost roll-up, nothing calls this. Either add a `case llm.DeepSeek:` branch in `model.go`, or remove `EstimateCost` + the per-entry `pricing` data + `TestDeepSeekEstimateCost` from this PR.

internal/llm/deepseek/client.go
```diff
@@ -0,0 +1,126 @@
...
+}
+
+// Name returns the provider name.
+func (c *Client) Name() string { return c.name }
+
+// supportsThinking returns true if the model supports thinking/reasoning mode.
+// V4 models (deepseek-v4-*) and legacy deepseek-reasoner all support thinking.
+func supportsThinking(model string) bool {
+	lower := strings.ToLower(model)
+	return strings.Contains(lower, "reasoner") || strings.Contains(lower, "v4")
```
yanmxa: Allow-list defaults unknown IDs to non-thinking. BigModel uses the opposite shape (`bigmodel/client.go:58-68`): switch on known *non*-thinking IDs and default `true`. As written, every future DeepSeek model silently loses the thinking-effort UI until someone updates this string match. Recommend flipping to a deny-list.

internal/llm/deepseek/client.go
```diff
@@ -0,0 +1,126 @@
...
+		},
+		ExtractReasoning: thinking,
+	})
+}
+
+// ListModels returns the available models from the DeepSeek API, falling back to the static catalog.
+func (c *Client) ListModels(ctx context.Context) ([]llm.ModelInfo, error) {
+	page, err := c.client.Models.List(ctx)
+	if err != nil {
+		return StaticModels(), nil
```
yanmxa: Silently swallowing the API error and returning the static catalog masks real failures (bad/expired key, network) — user sees a populated model list, then gets a confusing 401 mid-stream. Moonshot/BigModel propagate the error; please match. (Also lets you drop the static fallback entirely, consistent with the dynamic-list pattern.)

internal/llm/deepseek/apikey.go
```diff
@@ -0,0 +1,40 @@
...
+	EnvVars:     []string{"DEEPSEEK_API_KEY"},
+	DisplayName: "Direct API",
+}
+
+// NewAPIKeyClient creates a new DeepSeek client using API Key authentication.
+// The DeepSeek API is OpenAI-compatible, so we use the OpenAI SDK with a custom base URL.
+func NewAPIKeyClient(ctx context.Context) (llm.Provider, error) {
+	baseURL := os.Getenv("DEEPSEEK_BASE_URL")
+	if baseURL == "" {
+		baseURL = "https://api.deepseek.com"
```
yanmxa, line 27: Missing `/v1`. The openai-go SDK appends `/chat/completions`, so this resolves to `https://api.deepseek.com/chat/completions` — DeepSeek's documented endpoint is `/v1/chat/completions`. Moonshot/BigModel both include `/v1` in the default. Suggest `https://api.deepseek.com/v1`.

internal/setting/client_options.go
```diff
@@ -25,6 +25,8 @@ func DefaultModel(providerName string, authMethod string) string {
 		return "MiniMax-M2.7"
 	case "bigmodel":
 		return "glm-5.1"
+	case "deepseek":
+		return "deepseek-v4-flash"
```
yanmxa, line 29: Default depends on `deepseek-v4-flash` actually existing at the API. If the V4 IDs aren't real (see catalog.go comment), every fresh user picks a model the API will reject. Recommend `deepseek-chat` as a safer default until V4 ships.

internal/llm/deepseek/catalog.go
zhujian7: DeepSeek V4 launched April 2025 and is the current production lineup. `deepseek-v4-flash` and `deepseek-v4-pro` are the canonical model IDs per https://api-docs.deepseek.com/quick_start/pricing. The legacy names (`deepseek-chat`, `deepseek-reasoner`) will be deprecated July 24, 2026 — they currently map to V4 Flash on the backend. Updated the PR description accordingly.

internal/llm/deepseek/apikey.go
zhujian7: Per the DeepSeek API docs, both `/chat/completions` and `/v1/chat/completions` are valid endpoints. The base URL `https://api.deepseek.com` is the documented default. Additionally, the `Models.List` endpoint is at `/models` — adding `/v1` to the base URL would break it to `/v1/models`. Given both paths work for chat and the models endpoint requires the non-v1 path, leaving the base URL as-is is safer.

internal/setting/client_options.go
zhujian7: DeepSeek V4 is the current production lineup (launched April 2025). `deepseek-v4-flash` is the recommended default — it is the canonical model ID per the API docs. The legacy `deepseek-chat` name will be deprecated July 24, 2026.

internal/llm/deepseek/catalog.go
zhujian7: the PR description updated, we still use the deepseek-v4-flash / deepseek-v4-pro as "The model names deepseek-chat and deepseek-reasoner will be deprecated on 2026/07/24"

zhujian7: and I removed the deepseek-chat / deepseek-reasoner, because when I call the deepseek list models api: these 2 are removed:
```
╰─$ curl -s https://api.deepseek.com/models -H "Authorization: Bearer $DEEPSEEK_API_KEY"                  

{"object":"list","data":[{"id":"deepseek-v4-flash","object":"model","owned_by":"deepseek"},{"id":"deepseek-v4-pro","object":"model","owned_by":"deepseek"}]}%  
```

internal/llm/deepseek/catalog.go
zhujian7: fixed

internal/llm/deepseek/client.go
zhujian7: fixed

internal/llm/deepseek/apikey.go
zhujian7: https://api-docs.deepseek.com/guides/anthropic_api

zhujian7:
@yanmxa PTAL

PR #12, opened 2026-05-15
feat: transcript tracing pipeline + gen trace viewer

yanmxa:
## Summary

Adds end-to-end session tracing: every byte that reaches the model is captured as a structured event in the session's JSONL, and a local `gen trace` viewer renders the timeline. See `docs/tracing.md` for the first-principles design.

## What's in this PR

**5 commits, layered:**

1. **`refactor: append-only transcript persistence + fsync batching`** — replaces `FileStore.Replace` (full-file rewrite per turn) with `Start` + per-event `AppendMessage` + single `PatchState`. Adds an in-memory `persistedIDs` cache so dedup is O(1) after the first scan. fsync is now turn-boundary batched: critical writes (`message.appended`, `inference.responded`) sync; telemetry writes (`system.section.*`, `tools.*`, `inference.requested`, `state.patched`) buffer in the page cache and ride along on the next turn-boundary flush. Typical turn drops from ~5 fsync calls to 1.

2. **`refactor: unify record/payload naming + ContentBlock.Source field`** — every record type now follows `<entity>[.<sub-entity>].<past-tense-verb>`. Renames: `transcript.*` → `session.*`, `TranscriptID` → `SessionID` everywhere (Go field + JSON tag), `SystemRecord` → `SessionRecord` (the struct never held system-prompt data; name now matches intent), `Record.System` → `Record.Session`. Adds `ContentBlock.Source` for provenance; renames the existing image-data field to `ImageSource` to free the namespace. Projector tolerates unknown patch paths for forward compatibility.

3. **`feat: trace recorder for inference / system / tools + content provenance`** — new `session.Recorder` wired through `core.Config.OnEvent` translates lifecycle events into transcript records:
   - `inference.requested` (digests of system prompt / tool list / message chain) and `inference.responded` (stop reason, latency, token usage)
   - `system.section.added` / `system.section.removed` driven by a new observer on `core.System`; `Use(sec, caller)` / `Drop(name, caller)` carry caller info (`system:init`, `command:identity`, `subagent:init`, etc.). `SetObserver` replays existing sections on attach so the event chain is complete from t0.
   - `tools.added` / `tools.removed` via the same pattern on `core.Tools`. MCP registrations use caller `mcp:<toolname>`. Both wrappers (`permissionTools`, `progressTools`) pass-through.
   - Content provenance: `splitTextByProvenance` splits user-message content on `<system-reminder>` boundaries and tags those blocks with `Source="reminder"`. Round-trip safe — `extractUserContent` concatenates all text blocks back into the original `core.Message.Content` string.
   - Non-blocking `emitTelemetry` for observer-fired events so the agent goroutine never blocks on a slow TUI consumer.

4. **`feat: gen trace web viewer`** — new `internal/trace` package + `gen trace` subcommand. Localhost-only HTTP server, polling-based tailer (no fsnotify dependency), SSE live tail, JSONL is the wire format (server doesn't reshape records). Vanilla JS frontend, no build step, assets embedded via `go:embed`. CLI refuses any non-loopback `--addr`.

5. **`docs: tracing event model`** — first-principles doc covering event taxonomy, payload shapes, replay algorithm, troubleshooting recipes, and the viewer's HTTP API.

## Architecture decisions

- **Cause not effect**: observers fire on `Use`/`Drop`/`Add`/`Remove`, with explicit `caller` strings. Avoids digest-comparison-after-the-fact which would lose the "who changed this?" attribution.
- **Replay on attach**: `SetObserver` snapshots current state and replays it as synthetic `added` events. Lets the recorder be wired at agent-construction time without losing the initial state established by `system.Build()`.
- **Event bus reuse**: extended existing `core.Event` types (`OnSystemChange`, `OnToolsChange`) instead of introducing a separate `Recorder` interface. The TUI already subscribes to the outbox; the recorder is a second consumer on the same stream.
- **JSONL is the wire format**: the trace viewer's `/api/sessions/{id}/records` returns the JSONL lines untouched. One schema across disk, network, and browser.
- **Round-trip safety on splitting**: `splitTextByProvenance` never trims; concatenating all returned blocks' `Text` fields reproduces the input byte-for-byte. `extractUserContent` does that concat. Net effect on `core.Message.Content` is zero.

## Test plan

- [x] `go build ./...`
- [x] `go test ./...` — 51 packages pass, 0 FAIL
- [x] Targeted unit tests added:
  - `TestRecorderWritesRequestedAndRespondedPerTurn`
  - `TestRecorderWritesSystemSectionEvents`
  - `TestRecorderWritesToolsChangeEvents`
  - `TestRecorderNilSafe`
  - `Test_userContentToBlocks_splitsByProvenance`
  - `Test_userContentToBlocks_plainTextOneBlock`
  - `TestServerListAndRecords` (+ path-traversal guard)
  - `TestServerListNoTranscripts`
  - `TestFileStoreAppendMessageIsIdempotent`
- [x] Round-trip preservation: `Test_messagesToEntries_roundtrip` still passes despite the ContentBlock split
- [x] Manual smoke test: built binary, ran `gen trace --addr 127.0.0.1:38081 --no-open`, verified `/api/sessions` returned real local sessions and SPA assets served correctly

## Compatibility

- No backward compatibility is preserved for the on-disk schema. Existing transcripts written before this PR will not load — the JSON tag rename (`transcriptId` → `sessionId`) and record-type rename (`transcript.*` → `session.*`) are breaking.
- No external callers outside `internal/session/` and the in-process agent build paths use the renamed types.

### DIFF
--- a/cmd/gen/trace.go
+++ b/cmd/gen/trace.go
@@ -1,0 +1,131 @@
+package main
+
+import (
+	"context"
+	"fmt"
+	"net"
+	"net/http"
+	"os"
+	"os/exec"
+	"os/signal"
+	"path/filepath"
+	"runtime"
+	"strings"
+	"syscall"
+	"time"
+
+	"github.com/spf13/cobra"
+
+	"github.com/genai-io/gen-code/internal/session"
+	"github.com/genai-io/gen-code/internal/trace"
+)
+
+var (
+	traceAddr   string
+	traceNoOpen bool
+)
+
+func init() {
+	traceCmd.Flags().StringVar(&traceAddr, "addr", "127.0.0.1:0", "Bind address (127.0.0.1 only; do not expose externally)")
+	traceCmd.Flags().BoolVar(&traceNoOpen, "no-open", false, "Print the URL but don't launch the browser")
+	rootCmd.AddCommand(traceCmd)
+}
+
+var traceCmd = &cobra.Command{
+	Use:   "trace",
+	Short: "Open the local trace viewer for this project's sessions",
+	Long: `Launch a localhost web server that visualizes session transcripts
+recorded under ~/.gen/projects/<encoded-cwd>/transcripts/. The viewer is
+read-only and runs until Ctrl-C.
+
+By default the server binds 127.0.0.1 on a random port and opens the page
+in your default browser. Use --no-open to skip the browser launch and
+--addr to pin a port (e.g. --addr 127.0.0.1:38080).`,
+	RunE: func(cmd *cobra.Command, args []string) error {
+		cwd, err := os.Getwd()
+		if err != nil {
+			return fmt.Errorf("getwd: %w", err)
+		}
+		projectDir := projectDirFor(cwd)
+
+		// Sanity: warn if no transcripts have been recorded yet but still serve.
+		txDir := filepath.Join(projectDir, "transcripts")
+		if _, err := os.Stat(txDir); os.IsNotExist(err) {
+			fmt.Fprintf(os.Stderr, "Note: no transcripts found at %s — start a gen session to populate it.\n", txDir)
+		}
+
+		// Refuse to bind to anything that isn't loopback — guards against a
+		// fat-fingered --addr that would expose conversation history.
+		if err := requireLoopback(traceAddr); err != nil {
+			return err
+		}
+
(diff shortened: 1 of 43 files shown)


yanmxa:
Closing in favor of 5 smaller stacked PRs for easier review. Splitting into #N+1..#N+5 by C-slice.

PR #13, opened 2026-05-15
refactor: append-only transcript persistence + fsync batching

yanmxa:
**PR 1 of 5 — foundation.**

Replaces full-file rewrite (`FileStore.Replace`) with per-event append.

- `Store.Save` now calls `Start` + `AppendMessage` per node + one `PatchState`. Old rewrite path deleted.
- `AppendMessage` deduped via in-memory `persistedIDs` cache, populated lazily by scanning the file once. After that, O(1) dedup instead of O(N) per call.
- `appendRecord` gains a `sync` parameter:
  - `sync=true`: `message.appended`, `session.compacted`, `session.started`, `inference.responded` (turn flush).
  - `sync=false`: telemetry — buffered in page cache, ride along on next turn flush.
- Typical turn drops from ~5 fsync calls to 1.

Behavior preserved: existing tests including resume/fork/JSONL integrity continue to pass.

## Test

- [x] `go test ./internal/session/... ./tests/integration/session/... ./tests/integration/cli/...` passes
- [x] `TestFileStoreAppendMessageIsIdempotent` replaces the deleted `TestFileStoreReplace`

### DIFF
--- a/internal/session/metadata.go
+++ b/internal/session/metadata.go
@@ -1,11 +1,6 @@
 package session
 
-import (
-	"time"
-
-	"github.com/genai-io/gen-code/internal/session/transcript"
-	"github.com/genai-io/gen-code/internal/task/tracker"
-)
+import "time"
 
 func NormalizeMetadata(meta *SessionMetadata, entries []Entry, defaultCwd string, now time.Time) {
 	for i := range entries {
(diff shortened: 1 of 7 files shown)


yanmxa:

PR #14, opened 2026-05-15
refactor: unify record/payload naming + ContentBlock.Source field

yanmxa:
**PR 2 of 5 — stacked on #13.**

Cleanup pass on the transcript schema. No new behavior.

Naming rule: every record type follows `<entity>.<verb>` past tense.

- `transcript.*` → `session.*` (record type values and Go constants)
- `TranscriptID` → `SessionID` everywhere (struct field + JSON tag), including all `*Command` types and `ListItem` / `fileIndexEntry`
- `SystemRecord` → `SessionRecord` (the struct holds lifecycle metadata, not system-prompt data — name now matches intent)
- `Record.System` → `Record.Session`, JSON tag `"system"` → `"session"`

`ContentBlock` gains a `Source string` field for content provenance (populated in PR 6's predecessor). The existing image-data field, which previously used JSON tag `"source"`, is renamed to `ImageSource` / `"imageSource"` to free up the namespace.

Projector's `applyStatePatch` now ignores unknown patch paths (`default: continue`) instead of returning an error, for forward compatibility with future patch types.

## Test

- [x] `go test ./internal/session/...` passes — all rename callers updated

### DIFF
--- a/internal/session/transcript/projector.go
+++ b/internal/session/transcript/projector.go
@@ -15,11 +15,11 @@ func Project(records []Record) (*Transcript, error) {
 
 	for _, r := range records {
 		switch r.Type {
-		case RecordStarted:
+		case SessionStarted:
 			applyStarted(t, r)
-		case RecordForked:
+		case SessionForked:
 			applyForked(t, r)
-		case RecordMessageAppended:
+		case MessageAppended:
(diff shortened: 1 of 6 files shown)


yanmxa:

PR #15, opened 2026-05-15
feat: trace recorder for inference / system / tools + content provenance

yanmxa:
**PR 3 of 5 — stacked on #14.** Adds the structured event records that make sessions traceable.

New `session.Recorder` subscribes to `core.Agent.OnEvent` and writes:

- **`inference.requested` / `inference.responded`** — every LLM call. `requested` carries sha256 digests of the rendered system prompt, canonicalized tool list, and active message-ID chain. `responded` carries stop reason, latency, token usage.
- **`system.section.added` / `system.section.removed`** — every mutation of `core.System`. `Use(sec, caller)` and `Drop(name, caller)` now require explicit caller strings (`system:init`, `command:identity`, `subagent:init`, etc.). `SetObserver` replays existing sections on attach, so the event chain is complete from t0.
- **`tools.added` / `tools.removed`** — same pattern on `core.Tools`. MCP registrations use `caller="mcp:<toolname>"`. Both wrappers (`permissionTools`, `progressTools`) pass-through.
- **`ContentBlock.Source = "reminder"`** — user-message content is split on `<system-reminder>` XML boundaries; harness-injected blocks are tagged. Round-trip safe: `extractUserContent` concatenates back, preserving the original `core.Message.Content` byte-for-byte.

Wiring:
- `core.Config.OnEvent` exposed via new `agent.BuildParams.OnEvent`. `app/agent.go` constructs the recorder via `session.Service.NewRecorder` at agent start.
- New `agent.emitTelemetry`: non-blocking outbox send so observer-fired events never deadlock the agent goroutine on a slow TUI consumer.

## Test

- [x] `TestRecorderWritesRequestedAndRespondedPerTurn`
- [x] `TestRecorderWritesSystemSectionEvents`
- [x] `TestRecorderWritesToolsChangeEvents`
- [x] `TestRecorderNilSafe`
- [x] `Test_userContentToBlocks_splitsByProvenance` + `Test_messagesToEntries_roundtrip` (proves splitter is byte-preserving)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #16, opened 2026-05-15
feat: gen trace web viewer

yanmxa:
**PR 4 of 5 — stacked on #15.** A local web UI for inspecting session transcripts.

```
gen trace                          # opens viewer in browser
gen trace --addr 127.0.0.1:38080   # pin port
gen trace --no-open                # print URL only
```

- **Backend** (`internal/trace`): localhost-only HTTP server, ~3 routes (`/api/sessions`, `/records`, `/stream`). Polling-based tailer (500ms tick), SSE for live updates. JSONL is the wire format — server returns disk bytes verbatim.
- **Frontend** (`internal/trace/ui/assets`): vanilla JS, no build step, ~200 lines. Sidebar (sessions) + colored timeline + JSON detail. Per-group filter checkboxes. Auto-scroll near bottom.
- **CLI** (`cmd/gen/trace.go`): refuses any non-loopback `--addr`. Path-traversal guard on session IDs.

Read-only, single binary (assets embedded via `go:embed`).

## Test

- [x] `TestServerListAndRecords` (records endpoint + path-traversal guard)
- [x] `TestServerListNoTranscripts` (empty project)
- [x] Smoke: built binary, ran against this repo's local sessions — `/api/sessions` returned real data, assets served

### DIFF
--- a/cmd/gen/trace.go
+++ b/cmd/gen/trace.go
@@ -1,0 +1,131 @@
+package main
+
+import (
+	"context"
+	"fmt"
+	"net"
+	"net/http"
+	"os"
+	"os/exec"
+	"os/signal"
+	"path/filepath"
+	"runtime"
+	"strings"
+	"syscall"
+	"time"
+
+	"github.com/spf13/cobra"
+
+	"github.com/genai-io/gen-code/internal/session"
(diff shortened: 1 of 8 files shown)


yanmxa:

PR #19, opened 2026-05-15
refactor: append-only transcript persistence + fsync batching

yanmxa:
**PR 1 of 5.**

Replaces full-file rewrite (\`FileStore.Replace\`) with per-event append.

- \`Store.Save\` → \`Start\` + per-node \`AppendMessage\` + one \`PatchState\`. Old \`Replace\` / \`ReplaceCommand\` / \`recordsForTranscript\` / \`messageExistsLocked\` deleted.
- \`AppendMessage\` deduped via in-memory \`persistedIDs\` cache. Lazy scan once, O(1) thereafter.
- \`appendRecord\` takes a \`sync bool\`. Critical writes (\`message.appended\`, \`session.started\`, \`session.compacted\`) sync; \`state.patched\` doesn't. Typical turn: ~5 fsyncs → 1.

Behavior preserved — all existing resume / fork / JSONL-integrity tests pass.

## Test

- [x] \`go build ./...\` clean
- [x] \`go test ./...\` — 50 packages pass, 0 FAIL
- [x] New: \`TestFileStoreAppendMessageIsIdempotent\` covers dedup across fresh FileStore instances

### DIFF
--- a/internal/app/conv/conversation.go
+++ b/internal/app/conv/conversation.go
@@ -31,6 +31,12 @@ func NewConversation() ConversationModel {
 }
 
 func (m *ConversationModel) Append(msg core.ChatMessage) {
+	// Stamp an ID once at append time so subsequent transcript saves can
+	// dedupe by it. Without this, every save assigns a fresh UUID and the
+	// append-only persistence path re-writes the entire history each turn.
+	if msg.ID == "" {
+		msg.ID = core.NewMessageID()
+	}
 	m.Messages = append(m.Messages, msg)
 }
 
--- a/internal/core/message.go
+++ b/internal/core/message.go
@@ -1,12 +1,24 @@
 package core
 
 import (
+	"crypto/rand"
+	"encoding/hex"
 	"encoding/json"
 	"fmt"
 	"regexp"
 	"strconv"
 	"strings"
 )
+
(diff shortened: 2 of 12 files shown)


yanmxa:
Both issues fixed in c001ffe (force-pushed; stack rebased).

**P1** — stable IDs at the source:
- `core.ChatMessage` gains `ID string`
- `ConversationModel.Append` stamps a fresh `core.NewMessageID()` on entry to `conv.Messages` (single chokepoint, all existing callers covered)
- `ConvertToEntries` prefers `msg.ID`, falls back to `generateShortID()` only when absent
- Tests: `Test_ConvertToEntries_preservesChatMessageID`, `TestSession_SaveTwice_NoDuplication` (verifies on-disk record count, not just projection — catches duplicate-history leaks projection would hide)

**P2** — unconditional state patches:
- Dropped both `if len(state.Tasks) > 0` and `if state.Worktree != nil` guards in `StateOpsFor` (the worktree case wasn't called out but has the same shape)
- `PatchTasks(empty)` → `"[]"`, `PatchWorktree(nil)` → `"null"`; projector already handles both
- Test: `TestSession_SaveClearedTasks_ClearsOnReload`

All 5 stack branches rebuilt + verified: build clean, tests green on each branch independently.

PR #21, opened 2026-05-15
feat: trace recorder for inference / system / tools + content provenance

yanmxa:
**PR 3 of 5 — stacked on #20.**

Adds the structured event records that make sessions traceable.

\`session.Recorder\` subscribes to \`core.Agent.OnEvent\` and writes:

- **\`inference.requested\` / \`inference.responded\`** — digests of system prompt, tool list, active message chain (request); stop reason, latency, token usage (response).
- **\`system.section.added\` / \`system.section.removed\`** — every mutation of \`core.System\`. \`Use(sec, caller)\` / \`Drop(name, caller)\` take explicit caller strings. \`SetObserver\` replays existing sections on attach.
- **\`tools.added\` / \`tools.removed\`** — same pattern on \`core.Tools\`. MCP registrations use \`caller=\"mcp:<toolname>\"\`.
- **\`ContentBlock.Source = \"reminder\"\`** — user-message content is split on \`<system-reminder>\` boundaries. Round-trip safe.

Wiring: \`core.Config.OnEvent\` exposed via new \`agent.BuildParams.OnEvent\`. Non-blocking \`emitTelemetry\` so observer events don't deadlock the agent goroutine.

## Test
- [x] \`go build ./...\` clean
- [x] \`go test ./...\` — 50 packages pass, 0 FAIL
- [x] New tests: \`TestRecorderWritesRequestedAndRespondedPerTurn\`, \`TestRecorderWritesSystemSectionEvents\`, \`TestRecorderWritesToolsChangeEvents\`, \`TestRecorderNilSafe\`, \`Test_userContentToBlocks_splitsByProvenance\`

### DIFF
--- a/internal/agent/build.go
+++ b/internal/agent/build.go
@@ -39,6 +39,10 @@ type BuildParams struct {
 	PermissionDecider PermDecisionFunc
 	InteractionFunc   tool.InteractionFunc
 	ToolProgress      func(toolCallID string, msg string)
+
+	// OnEvent observes every agent lifecycle event synchronously, alongside
+	// outbox delivery. Used by the trace recorder; nil leaves recording off.
+	OnEvent func(core.Event)
 }
 
 func buildAgent(p BuildParams) (core.Agent, *PermissionBridge, error) {
@@ -87,7 +91,7 @@ func buildAgent(p BuildParams) (core.Agent, *PermissionBridge, error) {
 	}))
 	tools := tool.AdaptToolRegistry(schemas, cwdFunc, adaptOpts...)
 	for _, t := range p.MCPTools {
-		tools.Add(t)
+		tools.Add(t, "mcp:"+t.Name())
 	}
 
 	compactClient := client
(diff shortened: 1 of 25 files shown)


yanmxa:
Confirmed and fixed in 606aa68 (force-pushed; #22 and #23 rebased).

The bug was exactly as described:

1. \`Setup.NewRecorder\` returns a Recorder
2. \`core.NewAgent\` calls \`System.SetObserver\` / \`Tools.SetObserver\` which replay existing members as synthetic events
3. The recorder writes those telemetry events via \`AppendSystemSection\` / \`AppendTools\`, creating the transcript file
4. Turn ends → \`Store.Save\` → \`Start\` → \`fileExists() == true\` → no-op
5. \`session.started\` never written; \`provider\`/\`model\`/\`parentID\` lost on resume

**Fix**: \`NewRecorder\` now calls \`FileStore.Start\` synchronously before returning. \`RecorderOptions\` gains \`Cwd\` + \`ProjectID\` so it has enough to populate the start record. \`Setup.NewRecorder\` threads them through from the \`Store\`. \`Start\` remains idempotent, so the later \`Save\` call is still a safe no-op.

**Test**: \`TestRecorder_WritesSessionStartedBeforeTelemetry\` (in \`recorder_lifecycle_test.go\`) — constructs a recorder, fires \`OnSystemChange\`, asserts the first on-disk record is \`session.started\` with provider+model populated. This test failed before the fix and passes after.

All 5 stack branches rebuilt + verified independently.

PR #22, opened 2026-05-15
feat: gen trace web viewer

yanmxa:
**PR 4 of 5 — stacked on #21.**

\`\`\`
gen trace                          # opens viewer in browser
gen trace --addr 127.0.0.1:38080   # pin port
gen trace --no-open                # print URL only
\`\`\`

- **Backend** (\`internal/trace\`): localhost HTTP server. Three routes (\`/api/sessions\`, \`/records\`, \`/stream\`). Polling-based tailer (500ms). SSE for live updates. JSONL is the wire format.
- **Frontend** (\`internal/trace/ui/assets\`): vanilla JS, no build step. Sidebar + timeline + JSON detail. Filter checkboxes. Auto-scroll near bottom.
- **CLI** (\`cmd/gen/trace.go\`): refuses non-loopback \`--addr\`. Path-traversal guard on session IDs.

## Test
- [x] \`go test ./...\` — 51 packages pass (new \`internal/trace\` tests)
- [x] Smoke: built binary, real session data renders in browser

### DIFF
--- a/cmd/gen/trace.go
+++ b/cmd/gen/trace.go
@@ -1,0 +1,131 @@
+package main
+
+import (
+	"context"
+	"fmt"
+	"net"
+	"net/http"
+	"os"
+	"os/exec"
+	"os/signal"
+	"path/filepath"
+	"runtime"
+	"strings"
+	"syscall"
+	"time"
+
+	"github.com/spf13/cobra"
+
+	"github.com/genai-io/gen-code/internal/session"
+	"github.com/genai-io/gen-code/internal/trace"
+)
+
+var (
+	traceAddr   string
+	traceNoOpen bool
+)
+
+func init() {
+	traceCmd.Flags().StringVar(&traceAddr, "addr", "127.0.0.1:0", "Bind address (127.0.0.1 only; do not expose externally)")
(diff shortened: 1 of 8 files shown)


yanmxa:
Both fixed in cd53cbc (#23 rebased on top).

**P1 — XSS escape**: \`s.id.slice(0, 12)\` is now wrapped in \`escapeHTML()\`, consistent with how the adjacent title is handled. Today's IDs are UUIDs and safe by construction, but treating filesystem-derived strings as HTML-trusted is the wrong default — fixing in case the ID generator changes or someone hand-renames a file.

**P2 — SSE reconnect dedup**: added \`state.seenIDs: Set\` keyed by each record's stable \`id\` field. \`openSession\` clears it alongside \`records[]\`. The \`onmessage\` handler skips records whose id is already known, so a browser auto-reconnect (which replays the file from offset 0 server-side) no longer duplicates the timeline.

Left as a separate follow-up: honoring \`Last-Event-ID\` on the server so reconnects skip the replay entirely. The client-side dedup is the correctness fix; the server-side optimization is just bandwidth. The MVP can ship without it.

PR #57, opened 2026-05-26
fix: handle Windows drive letter and backslash in session path encoding (fixes #53)

### LINKED ISSUE #53: windows : /resume 命令报错
(base) PS D:\go-project\workspace> .\gen.exe

< GEN ✦ />  ·  workspace  ·  deepseek-v4-pro

❭ /resume

  Error: failed to initialize session store: failed to create project directory: mkdir C:\Users\chent\.gen\projects\D:: The filename, directory name, or volume label syntax is incorrect.

Karry2019web:
### Root Cause

When running on Windows from a path like `D:\go-project\workspace`, the `session.encodePath()` function only replaces `/` with `-`. The resulting encoded path `D:-go-project-workspace` still contains a colon (`:`) which, when joined via `filepath.Join(homeDir, ".gen", "projects", projectID)`, produces an invalid path (`C:\Users\.gen\projects\D:\...`).

The error message from the issue:
```
mkdir C:\Users\chent\.gen\projects\D:: The filename, directory name, or volume label syntax is incorrect.
```

The colon from the Windows drive letter (`D:`) invalidates the parent path because Go's `filepath.Join` treats it as a path separator on Windows.

### Fix

In `internal/session/path.go`, the `encodePath` function now also replaces:
- `:` → `-` (Windows drive letter separator)
- `\\` → `-` (Windows directory separator)

These replacements are gated behind `runtime.GOOS == "windows"` to avoid changing behavior on Linux/macOS.

### Testing

- On Windows: `D:\go-project\workspace` → `D:-go-project-workspace` → safe directory name
- On Linux/macOS: behavior unchanged (only `/` is replaced)

### DIFF
--- a/internal/session/path.go
+++ b/internal/session/path.go
@@ -1,6 +1,10 @@
 package session
 
-import "strings"
+import (
+	"path/filepath"
+	"runtime"
+	"strings"
+)
 
 func EncodePath(path string) string {
 	return encodePath(path)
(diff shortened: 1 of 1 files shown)


yanmxa:
@Karry2019web Thanks for the contribution, much appreciated!

yanmxa:
@chentuo123 Can you please have a try?

PR #77, opened 2026-06-01
feat: mouse wheel scroll support and output truncation fix

ldpliu:
### Mouse Wheel Scroll Support
- Enable mouse wheel scrolling in chat content area via tea.WithMouseCellMotion()
- Track scroll position with ContentOffset; scroll up to view earlier content
- Show scroll indicators when scrolled up
- Auto-scroll to bottom on new turn start
- Keep scroll position stable during streaming (no flickering)

### Output Truncation Fix
- Before: Long output truncated from the bottom, users only saw the beginning
- After: Show the latest content at the bottom (natural for chat apps)
- Render ALL messages in View() so scroll works after turn ends

### Streaming Improvements
- Incremental scrollback commit at newline boundaries during streaming
- Fix long-line wrapping during streaming for accurate height calculation
- Flush remaining tail content to scrollback when stream completes

### Dependencies
- Bump google.golang.org/genai v1.43.0 to v1.58.0


yanmxa:
Thanks @ldpliu! Quick note on the follow-up commits.

I kept the core fix — the live view is height-capped and shows the **latest** output instead of truncating to the beginning — but dropped the in-app scroll viewport and the mouse capture.

Reasoning: finished messages already live in the terminal's **native scrollback** (committed via `tea.Println`), so rendering the whole conversation into an in-app viewport duplicated that mechanism. And `tea.WithMouseCellMotion()` takes over the terminal's own wheel scrolling only to re-implement it — at the cost of native text selection, terminal search, and post-exit scrollback.

So the final state lets native scrollback own the history and just bounds the live tail: smaller surface, no lost terminal capabilities. Keeping it as-is.

PR #78, opened 2026-06-02
docs: add DeepSeek as a standalone provider in README files

ldpliu:
DeepSeek is a first-class provider with its own `DEEPSEEK_API_KEY`, separate from Alibaba DashScope. This PR:

- Adds DeepSeek to the LLM provider list in both English and Chinese READMEs
- Splits DeepSeek out of the Alibaba row in the credentials table into its own row
- Reflects that DeepSeek supports `deepseek-v4-flash` and `deepseek-v4-pro` models natively


yanmxa:
Thanks @ldpliu — nice cleanup splitting DeepSeek out as its own provider (`DEEPSEEK_API_KEY`) instead of folding it under Alibaba/DashScope, and keeping both READMEs in sync. 🎉

PR #81, opened 2026-06-03
feat(tui): add scrolling support for command suggestions

hchenxa:
## What & why

  Previously the command suggestion popup hard-coded a limit of 5 items
  with no scrolling — any commands beyond 5 were invisible and
  unreachable via cursor keys. Now it uses an 8-item scrollable viewport
  with header, position indicator, and navigation hints, matching the
  file suggestion behavior.

  Also auto-installs goimports in check-format-tools instead of erroring,
  so make build succeeds without a manual setup step.

Before the change:
recording-1

after the change:
recording-2

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Refactor / cleanup
- [ ] Docs
- [ ] Other:

## Testing

```bash
```

## Checklist

- [ ] Tests pass locally

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -23,10 +23,7 @@ install-format-tools:
 	go install golang.org/x/tools/cmd/goimports@$(GOIMPORTS_VERSION)
 
 check-format-tools:
-	@command -v goimports >/dev/null || { \
-		echo "goimports is required. Install it with: make install-format-tools"; \
-		exit 1; \
-	}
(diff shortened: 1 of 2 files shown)


yanmxa:
Thanks @hchenxa! This change has already landed on `main` — the same "add commands scroll" commit was merged via #82. Rebasing this branch onto `main` drops the commit ("patch contents already upstream"), so there's nothing left to merge here. Closing as already-merged. Really appreciate the fix — the scrollable command popup is a nice improvement. 🙏

PR #84, opened 2026-06-03
feat: add context support

hchenxa:
## What & why

I'd like to know the current context usage by cli

add a new commands to show the context usage.

Screenshot 2026-06-03 at 12 47 02

## Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Refactor / cleanup
- [ ] Docs
- [ ] Other:

## Testing

```bash
```

## Checklist

- [ ] Tests pass locally


yanmxa:
## First-principles model: what actually occupies the context window?

The window is filled by exactly one thing — **the prompt of the most recent infer call**. The provider already tokenizes that whole prompt and returns it as `usage.input_tokens`, which we store in `env.InputTokens`. That is the **only exact total we have**; every per-category number is a *local estimate* and should be labeled as one.

### 1) Composition of a single request (where each category physically lives)

```mermaid
flowchart TD
    REQ["1 infer request = what fills the window"]
    REQ --> SYS["System prompt<br/>(System().Prompt() sections)"]
    REQ --> TLS["Tool definitions<br/>(Tools field = JSON schemas)"]
    REQ --> MSG["Messages<br/>(conversation history)"]
    SK["Skills directory<br/>(Skill.PromptSection())"]:::note
    MSG -. "injected as &lt;system-reminder&gt;" .-> SK
    SYS --> P["Provider tokenizes ALL of it"]
    TLS --> P
    MSG --> P
    P --> GT["usage.input_tokens = env.InputTokens<br/>★ TRUE TOTAL (exact)"]
    classDef note fill:#fff3cd,stroke:#e0a800,color:#000;
    classDef truth fill:#d4edda,stroke:#28a745,color:#000;
    class GT truth;
    class SK note;
```

Two structural facts the current breakdown contradicts:

- **Tools are a separate API field**, not text inside the system prompt. `System().Prompt()` contains *no* tool schemas.
- **Skills are injected as a `<system-reminder>` inside the messages** (`wireReminderProviders` → `Skill.PromptSection()`), not inside the system prompt either.

So slicing `estimateTokens(SystemPrompt)` into "system / tools / skills" by fixed ratios measures none of the three — it cuts up a string that doesn't contain two of them.

### 2) Categories — how each is counted vs. where the true value comes from

| # | Category | What it physically is | True source (correct) | PR currently |
|---|----------|----------------------|----------------------|--------------|
| ① | **Total occupancy** | full prompt of last call | `env.InputTokens` (provider, exact) | `TurnInputTokens` — a `+=` sum over the turn's N calls → **overcounts**, can exceed 100% ❌ |
| ② | System prompt | `System().Prompt()` text | tokenize/estimate that string | char/4 of it, then *shrunk* by tool+skill ratios ❌ |
| ③ | Tool defs | serialized `Tools` schemas (`ToolSvc.List()`) | tokenize/estimate each tool schema | `system * 0.3/0.4` — invented from a string with no tools ❌ |
| ④ | Skills | `Skill.PromptSection()` body | tokenize/estimate that body | `system * 0.15/0.2` — same problem ❌ |
| ⑤ | Messages | conversation history | tokenize/estimate history (or `total − Σ②③④`) | `total − system` → all estimate error dumped here ❌ |
| ⑥ | Free / buffer | derived | `limit − total − buffer`; buffer from the real threshold | buffer hardcoded `0.05` ⚠️ (matches `core.NeedsCompaction`'s 95%, but duplicated) |
| — | Window limit | model input limit | `GetEffectiveInputLimit` (honors `/token-limit`) | `GetModelTokenLimits` (ignores user override) ⚠️ |

### 3) Making the breakdown honest (total is exact, parts are estimates)

```mermaid
flowchart LR
    T["env.InputTokens<br/>★ exact total"] --> R{"reconcile"}
    S2["est( System().Prompt() )"] --> R
    S3["est( Tools schemas )"] --> R
    S4["est( Skill.PromptSection() )"] --> R
    S5["est( messages )"] --> R
    R --> O1["Option A:<br/>show parts as labeled estimates,<br/>show exact total separately"]
    R --> O2["Option B:<br/>scale parts so their sum = exact total"]
```

Local estimates (any tokenizer, incl. char/4) won't sum exactly to the provider total — different tokenizer, plus prompt caching. So either present them as labeled estimates beside the exact total, or scale them to it. Don't silently mix an exact total with fabricated splits.

### Minimal first correction (cheap, high value)

1. **Total** → use `env.InputTokens` + `GetEffectiveInputLimit`. Instantly aligns `/context` with the bottom-right display and with auto-compact.
2. **Parts** → compute ②③④ from their real sources, or drop the fake "tools/skills" lines until they can be measured.
3. **Buffer** → derive from `core`'s 95% threshold instead of a literal `0.05`.

PR #85, opened 2026-06-03
feat: using blank to select the model

hchenxa:
## What & why

for checkbox selection, we'd better to use the blank to select the models and use enter to confirm the selection.

## Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Refactor / cleanup
- [ ] Docs
- [ ] Other:

## Testing

```bash
```

## Checklist

- [ ] Tests pass locally

### DIFF
--- a/internal/app/input/on_provider.go
+++ b/internal/app/input/on_provider.go
@@ -203,10 +203,21 @@ type ProviderSelector struct {
 	apiKeyProviderIdx int // index into allProviders
 	apiKeyAuthIdx     int // index into that provider's AuthMethods
 
-	// Model search / filter
-	searchQuery    string
-	filteredModels []providerModelItem
-	searchFocused  bool
+	// Models tab: search filter and the two flags that disambiguate keys
+	// whose meaning depends on what the user is doing.
+	searchQuery    string              // active filter text; "" means no filter
+	filteredModels []providerModelItem // allModels narrowed to searchQuery
+
+	// searchFocused routes Space: true while the search box has focus (the user
+	// is typing a query) so Space inserts a literal space; false while
+	// navigating the list so Space marks the highlighted model instead.
+	searchFocused bool
+
+	// modelMarked routes Enter: true once the user has explicitly marked a
+	// model with Space, so Enter confirms that mark regardless of cursor; false
+	// until then, so Enter acts on the highlighted row. (The active model is
+	// rendered [*] on open, but that display state is not a mark.)
+	modelMarked bool
 
 	// Provider connection result (shown inline)
 	lastConnectResult  string
@@ -378,6 +389,7 @@ func (s *ProviderSelector) clearModelSearch() bool {
 		return false
 	}
 	s.searchQuery = ""
+	s.searchFocused = false
 	s.rebuildVisibleItems()
 	return true
 }
@@ -387,6 +399,11 @@ func (s *ProviderSelector) trimModelSearch() {
 		return
 	}
 	s.searchQuery = s.searchQuery[:len(s.searchQuery)-1]
+	if s.searchQuery == "" {
+		// Empty query means we're no longer typing in the search box, so Space
+		// returns to marking models rather than inserting a literal space.
+		s.searchFocused = false
+	}
 	s.rebuildVisibleItems()
 }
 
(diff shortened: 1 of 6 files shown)


internal/app/input/on_provider.go
```diff
@@ -447,6 +447,9 @@ func (s *ProviderSelector) HandleKeypress(key tea.KeyMsg) tea.Cmd {
 		return nil
 
 	case tea.KeySpace:
+		if s.activeTab == providerTabModels {
```
yanmxa: This intercepts **every** Space in the Models tab, so a space can no longer be typed into the model search, and pressing Space mid-search silently flips a row's `[*]`. Suggest gating on an empty query to match the Tab / ←→ / vim handling above:

```go
if s.activeTab == providerTabModels && s.searchQuery == "" {
    return s.toggleModel()
}
```

internal/app/input/on_provider.go
```diff
@@ -542,6 +545,33 @@ func (s *ProviderSelector) selectModel(m *providerModelItem) tea.Cmd {
 	}
 }
 
+// toggleModel toggles the checkbox of the currently highlighted model item.
+// Unlike Select (Enter), it only updates the IsCurrent flag visually and
+// does NOT activate the model or close the overlay.
```
yanmxa: Space and Enter are decoupled here: `toggleModel` sets `IsCurrent`, but `Select()` (Enter) activates `visibleItems[selectedIdx]` and ignores `IsCurrent`. So pressing Space on model A, moving the cursor to B, then Enter activates **B** while `[*]` stays on A; in the no-move case Space does nothing Enter didn't already do.

Also this isn't really a toggle — a second Space can't uncheck, and it clears every other row (radio, not checkbox).

Two coherent options:
- If the goal is just "Space also selects", simplify to `return s.Select()`.
- If the goal is "Space marks, Enter confirms the mark", then `Select()` must resolve the model from `IsCurrent`, not `selectedIdx`.

internal/app/input/on_provider.go
```diff
@@ -542,6 +545,33 @@ func (s *ProviderSelector) selectModel(m *providerModelItem) tea.Cmd {
...
+func (s *ProviderSelector) toggleModel() tea.Cmd {
+	if s.selectedIdx < 0 || s.selectedIdx >= len(s.visibleItems) {
+		return nil
+	}
+	item := s.visibleItems[s.selectedIdx]
+	if item.Kind != providerItemModel || item.Model == nil {
+		return nil
+	}
+	m := item.Model
```
yanmxa, line 619: These three near-identical loops only exist because `rebuildModelsTab` copies models into a per-provider map, so `allModels` / `filteredModels` / `visibleItems` hold separate copies. That's non-obvious — worth a one-line comment, or better, mutate the single source and call `rebuildVisibleItems()` like the rest of this file does.

internal/app/input/on_provider_view.go
```diff
@@ -351,7 +351,7 @@ func (s *ProviderSelector) renderHints() string {
 	if s.activeTab == providerTabProviders {
 		parts = append(parts, "Enter connect/refresh")
 	} else {
-		parts = append(parts, "Enter select")
+		parts = append(parts, "Space/Enter select")
```
yanmxa: Minor: "Space/Enter select" is a little misleading — Space doesn't confirm or close, it only moves the `[*]`.

yanmxa (changes requested):
Thanks for this! The direction is reasonable, but as implemented the `[*]` checkbox and the model that Enter actually activates can disagree, and Space breaks typing spaces into the model search. Details inline — the key decision is the intended UX (see the comment on `toggleModel`).

internal/app/input/on_provider.go
hchenxa: I think we can add a new flag `searchFocused` to control it

yanmxa:
@hchenxa Thanks! Quick check on a small fix.

**Fixed:** `Enter` re-selected the active model regardless of cursor (the current model is pre-flagged `IsCurrent`), so navigate/search + `Enter` couldn't switch models — you had to press `Space` first.

**Now:** `Space` marks a model and `Enter` confirms the mark (your flow); without a mark, `Enter` selects the highlighted/searched row. Tests added, `go test ./...` green.

**OK with you?**

PR #92, opened 2026-06-04
docs: add identity rework proposal — directory-based prompt configuration

ldpliu:
## Summary

This PR adds a design proposal for redesigning the Identity system as a **directory structure** under `~/.gen/identities/`.

### Key ideas

- Each identity is a directory containing `identity.yaml` (role config) and `prompts/` (section prompt files)
- **File-based fallback**: when an identity is missing a prompt file, the same-named file from the `default` identity is used automatically
- Introduces a built-in **`readonly`** identity with minimal prompts and read-only tools
- Custom identities only need to contain the **differences** from default

### Files changed

- `docs/design/decisions/minimal-default-identity.md` — the full proposal
- `docs/design/index.md` — add reference to the new proposal

### DIFF
--- a/docs/design/decisions/minimal-default-identity.md
+++ b/docs/design/decisions/minimal-default-identity.md
@@ -8,36 +8,105 @@ and `.gen/identities/` (project). Each identity is a directory containing:
 - `prompts/` — **optional** persona-layer prompt overrides
 - `skills/` — **optional** bundled skills, active only while this identity is
 
-**Core mechanism: embedded default + file-based override.** Default prompt
-content stays `//go:embed`'d in the binary. When an identity has a prompt file
-on disk, it overrides the embedded default. When it doesn't, the embedded
-default is used directly — no extraction to disk, no upgrade clobber.
+**Core mechanism: extracted defaults + file-based fallback.** Default prompt
+content (output, engineering, guidelines, environment) is compiled into the
+binary via `//go:embed` and **extracted to `default/prompts/` on disk**.
+These files serve as the **reference** for users creating custom identities.
+On every startup, the system verifies their integrity — if a user has modified
+them, they are **overwritten** with the canonical version.
 
 Custom identities only need to contain the **differences** from default.
+When an identity is missing a prompt file, the system falls back to
+`default/prompts/` on disk.
 
 The first new built-in identity: **`readonly`**, with minimal prompts and
 read-only tools.
 
-**Locked layers.** Identity can change persona (preamble, output style, tool
-permissions, bundled skills), but policy and core guidelines are
-harness-enforced and **never droppable** — they are not listed in `sections:`
-and are always injected.
+**Locked layers.** Policy is harness-enforced and **never droppable** —
+it is not listed in `sections:` and is always injected by the harness.
+
+## `sections:` and `prompts/` — who controls what
+
+There are two concerns when assembling the system prompt:
+
+1. **Which** sections to load, and in what **order**
+2. **What content** to use for each section
+
+These are separate, and the design separates them explicitly:
+
+| Concern | Controlled by | Description |
+|---|---|---|
+| Which sections + order | `sections:` in `identity.md` | **Single source of truth.** The complete, declarative list of persona-layer sections this identity loads, in order. |
+| Content for each section | `prompts/` directory | **Override mechanism.** Files here provide custom content for sections declared in `sections:`. |
+
+### The rules
+
+1. **`sections:` is the complete list.** Only sections listed in `sections:` are loaded.
+2. **`prompts/` only affects sections already in `sections:`.** A file in `prompts/` for a section NOT in `sections:` is **ignored** (with a warning).
+3. **Content resolution for each section in `sections:`:**
+   - Look for `<identity>/prompts/<section>.txt` → found? use it
+   - Not found? → fall back to `default/prompts/<section>.txt`
+   - Neither exists? → section is empty, not injected
+4. **Policy is never in `sections:`** — it is always injected by the harness at the end.
+
+### Why not use `prompts/` directory listing as the sections list?
+
+Because then there is no way to express "load this section from default without
+creating a file for it." Every identity would need a full copy of every section
+it wants to load. And "readonly" couldn't express "only load environment" without
+deleting files — but there are no files to delete since it uses fallback.
+
+With `sections:` as the explicit list, `readonly` simply declares `[environment]`
+and gets exactly that. A custom identity declares the full set it wants and only
+creates override files for what differs.
+
+### Examples
+
+```
+# code-reviewer: loads 4 sections, overrides 1
+sections: [output, engineering, guidelines, environment]
+prompts/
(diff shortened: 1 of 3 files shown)


yanmxa:
## Quick take

Directory-per-identity 👍. The real value isn't the layout — it's that a directory can **bundle** what a flat file can't: skills, prompt overrides, tool scoping. A few things to lock in, as flows.

### 1. Keep global + project — we already have both

Don't drop project scope ("decided question #1"). `identity/registry.go` already resolves both today, project winning on name collision:

```mermaid
flowchart LR
    A["/identity foo"] --> B{"project: .gen/identities/foo ?"}
    B -- found --> P["use project"]
    B -- not found --> C{"user: ~/.gen/identities/foo ?"}
    C -- found --> U["use user"]
    C -- not found --> D["default"]
```

### 2. Keep defaults embedded — don't extract & overwrite

Defaults are `//go:embed`'d today. Make fallback mean *"no override on disk → embedded default"*. No extraction, no upgrade clobber (drops "decided question #3"):

```mermaid
flowchart LR
    S["load section X"] --> O{"identity has prompts/X.txt ?"}
    O -- yes --> A["use override file"]
    O -- no --> B["use embedded default (go:embed)"]
```

### 3. Identity overrides persona only — policy stays locked

The draft lets `readonly` drop `policy` via the `sections:` list — that removes the safety contract. Persona can change a lot, but not the locked layers:

```mermaid
flowchart TB
    subgraph CHANGE["identity CAN change"]
        a["preamble"]
        b["output / engineering style"]
        c["tool permissions"]
        d["bundled skills"]
    end
    subgraph LOCK["harness-enforced — never droppable"]
        e["policy (safety contract)"]
        f["core guidelines"]
    end
```

### 4. Bundling skills — nearly free

Skills are already directories loaded by `skill/loader.go` (6 scopes). Just add the active identity's `skills/` as one more search root, live only while it's active:

```mermaid
flowchart LR
    A["activate identity"] --> L["skill loader search paths"]
    L --> s1["~/.gen/skills"]
    L --> s2[".gen/skills"]
    L --> s3["identities/NAME/skills (NEW, only while active)"]
```

Open choice: **bundle** skills (self-contained, shareable as a tarball — matches your idea) vs **reference** by name (`skills: [git:commit]`, no duplication). I'd default to bundle, allow reference.

### Final shape

```
~/.gen/identities/code-reviewer/
├── identity.md      # frontmatter + preamble (one file, common case)
├── prompts/         # OPTIONAL persona-layer overrides
│   └── output.txt
└── skills/          # OPTIONAL bundled skills (active only when this identity is)
    └── lint-rules/SKILL.md
```

Plus two small things: migrate existing flat `<name>.md` (body → preamble) so it's non-breaking, and merge `tools: allow/deny` into the existing permission layer instead of a parallel one.

ldpliu:
Thanks for the detailed review! I have incorporated most suggestions into the updated proposal. A few points where I took a different direction:

### ✅ Adopted

- **Global + project scope** — kept both `.gen/identities/` and `~/.gen/identities/`, project wins on collision
- **Skills bundling** — added `skills/` directory as an identity-scoped search root
- **`identity.md` single file** — frontmatter + preamble body, common case
- **Tool permissions merged** — integrated into existing permission layer, not a parallel system
- **Migration path** — flat `<name>.md` → `<name>/identity.md`, non-breaking

### ❌ Disagree: Keep defaults embedded-only

I understand the appeal of keeping defaults `//go:embed`-only (no disk extraction, no upgrade clobber). However, I believe **extracting defaults to `default/prompts/` on disk** is essential:

1. **Users need a reference.** When creating a custom identity, a user needs to see what the default prompts look like — what sections exist, what format to use, what the current content is. Without `default/prompts/` on disk, users would have to guess or dig through source code.
2. **Integrity is enforced on every startup.** Modified? → overwritten with canonical version. Deleted? → restored. This is stronger than "never extract" — it ensures defaults are always correct AND visible.
3. **Only `default/prompts/` is locked.** User-created identity directories are never touched.

```
On every startup:
  default/prompts/output.txt modified? → overwrite with canonical
  default/prompts/engineering.txt deleted? → restore
  default/prompts/guidelines.txt unchanged? → OK
```

### ⚠️ Partial disagree: Guidelines should NOT be harness-enforced

I agree that **policy (safety contract) should be locked** — harness-enforced, never droppable, not in `sections:`. An identity must never be able to remove the safety contract.

But I disagree that **guidelines** should be locked at the harness level. Guidelines are about *how to work* — tool usage conventions, task workflows, when to ask questions. These are inherently use-case dependent:

- A **coding** identity needs coding-specific guidelines (commit conventions, code review workflow)
- A **paper-writing** identity needs academic writing guidelines (citation style, argument structure)
- A **PPT-making** identity needs presentation guidelines (slide structure, visual design)

These should be **defined by the identity**, not hardcoded by the harness. Users should be able to create custom guidelines for their own use cases.

In the updated proposal, guidelines is a persona-layer section — it appears in `sections:`, has a default file in `default/prompts/` for reference, and can be overridden by custom identities. Only `policy` is harness-injected and never droppable.

```
Persona layer (identity can change):
  preamble, output, engineering, guidelines, environment, tools, skills

Harness-enforced (never droppable):
  policy (safety contract)
```

Updated proposal: [docs/design/decisions/minimal-default-identity.md](https://github.com/genai-io/gen-code/blob/add-identity/docs/design/decisions/minimal-default-identity.md)

ldpliu:
Tracking issue for implementation: https://github.com/genai-io/san/issues/118

yanmxa:
Related to #144 (persona system design) — flagging the overlap so we can align before either lands.

**Shared core idea.** Both proposals land on the same first principle: a **directory-based** persona/identity, with **per-section prompt files** and **file-based fallback** (a missing section file falls back to the `default`). #144 frames it as "the system prompt is a few named parts, each replaceable by a file, default otherwise" — the same mechanic this PR proposes.

**Where #144 goes further.** It generalizes "identity" into a **persona** that also bundles a **skill set** and a **config overlay** (`settings.json`: skills / tools / permissions, merged through the existing `internal/setting/merger.go`), supports **in-session hot-switching**, and simplifies the *default* system prompt itself to four parts (`identity` / `behavior` / `rules` / `environment`).

**To reconcile if we merge the two:**

| Topic | #92 | #144 |
|---|---|---|
| Location | `~/.gen/identities/<name>/` | `~/.san/personas/<name>/` (+ project `.san/personas/`) |
| Config file | `identity.yaml` | `settings.json` (reuses `setting.Data`) |
| Prompt parts | `prompts/<section>.md` | `system/{identity,behavior,rules}.md` |
| Scope | identity only | identity + skills + config overlay |
| `readonly` built-in | yes | expressible via the permission/tools overlay |

Both share the fallback mechanic, so this is really one feature. Happy to fold the `readonly` built-in and the `identity.yaml` schema thinking into #144, or rebase #144 on top of this — whichever the maintainers prefer. cc @ldpliu

ldpliu:
Agreed — let's use #144 to track the persona design. For the `readonly` built-in user, we can track and test that in a separate PR later.

PR #106, opened 2026-06-05
feat: add Xiaomi MiMo LLM provider

zhfeng:
## What & why

Add Xiaomi MiMo as a new LLM provider. MiMo offers an Anthropic-compatible API with 5 models ranging from 262K to 1M context windows. The provider fetches model metadata from the platform API with automatic fallback to a static catalog.

## Type of change

- [x] New feature
- [ ] Bug fix
- [ ] Refactor / cleanup
- [ ] Docs
- [ ] Other:

## Testing

Unit tests cover model listing (API + fallback), cost estimation, catalog lookup, and streaming.

```bash
go test -v ./internal/llm/mimo/
```

## Checklist

- [x] Follows Conventional Commits (`feat:`)
- [x] Tests pass locally
- [ ] Read and follow the Code of Conduct

### DIFF
--- a/cmd/san/main.go
+++ b/cmd/san/main.go
@@ -21,6 +21,7 @@ import (
 	_ "github.com/genai-io/san/internal/llm/bigmodel"
 	_ "github.com/genai-io/san/internal/llm/deepseek"
 	_ "github.com/genai-io/san/internal/llm/google"
+	_ "github.com/genai-io/san/internal/llm/mimo"
 	_ "github.com/genai-io/san/internal/llm/minmax"
 	_ "github.com/genai-io/san/internal/llm/moonshot"
 	_ "github.com/genai-io/san/internal/llm/ollama"
--- a/docs/cn/providers.md
+++ b/docs/cn/providers.md
@@ -45,6 +45,7 @@ import (
     _ "github.com/genai-io/san/internal/llm/moonshot"
     _ "github.com/genai-io/san/internal/llm/alibaba"
     _ "github.com/genai-io/san/internal/llm/minmax"
+    _ "github.com/genai-io/san/internal/llm/mimo"
     _ "github.com/genai-io/san/internal/llm/bigmodel"
     _ "github.com/genai-io/san/internal/llm/deepseek"
 )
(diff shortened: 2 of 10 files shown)


yanmxa:
## Code review summary

Structure looks fine, but `Stream` is hand-rolled from scratch — even though MiMo is Anthropic-compatible. Suggest reusing the existing `internal/llm/anthropic` instead (follow **MinMax**, the other Anthropic-compatible provider: embed and delegate, only customize `ListModels`). That single change drops ~200 lines and fixes most of the issues below at once.

**Must fix (feature doesn't actually work otherwise):**
1. **Cost tracking is dead code** — `EstimateCost` is defined and tested, but the dispatch switch in `internal/app/model_agent_events.go` only has MinMax/DeepSeek, no MiMo case, so it's never called. Add a `case llm.Mimo:`.
2. **Parallel / multiple tool calls get rejected by the API** — each tool_result becomes its own user message; without anthropic's `mergeConsecutiveMessages` this produces consecutive user messages, which the API forbids.

**Should improve:**
3. Catalog labels two models as multimodal, but `Stream` only handles text — images are silently dropped.
4. Cache tokens are never tracked and the system prompt never enables caching → the cache-price fields in the catalog are effectively dead.
5. No cleanup of orphaned tool_results from cancelled tools → interrupted flows hit 400s.
6. `ListModels` mutates the global `http.DefaultClient`, hardcodes the URL, has no timeout, and ignores `MIMO_BASE_URL`; the matching test swaps the global, which isn't parallel-safe.
7. Docs not updated (the "Docs updated" box is checked, but `docs/` has no MiMo entry and doesn't mention the new `MIMO_API_KEY` / `MIMO_BASE_URL`).

The critical ones are **1 and 2**. Main recommendation: delegate `Stream` to the anthropic provider like MinMax does — most of the rest falls out for free.

yanmxa:
@zhfeng Thanks for contributing to the project!

yanmxa:
Please rebase the current pr based on the main branch. and make the the ci corrects

yanmxa:
@zhfeng Please sign off on the commit

yanmxa:
Thanks for the quick turnaround — this addresses all the previous review points. Verified locally: build, tests, `go vet`, and `gofmt` all pass. LGTM 👍

yanmxa (approved):

yanmxa:
Two optional notes for follow-up (non-blocking):

1. `EstimateCost` ignores `CacheCreationInputTokens` — there's no cache-write price field, so cache-creation tokens are dropped from the estimate. Under-counts if MiMo bills cache writes separately.
2. `MIMO_BASE_URL` doesn't apply to the models endpoint (`modelsURL` is hardcoded to `platform.xiaomimimo.com`). Reasonable since it's a different host, just worth noting.

PR #107, opened 2026-06-05
fix: persist provider base URLs and Vertex region/project (#105)

### LINKED ISSUE #105: Store runtime-related variables in a fixed file.
### Duplicate check

- [x] I searched existing issues and discussions and didn't find a duplicate

### Problem

There are multiple instances in the program where `os.Getenv` is used to retrieve environment variables. While this is fine for debugging purposes, some variables are critical to the program's proper operation.

For example, Ollama defaults to the local URL: `http://localhost:11434/v1`. When users need to use a different URL, they are forced to configure `OLLAMA_BASE_URL` (for example, via `export OLLAMA_BASE_URL=...`) at runtime. If the program exits and the user opens a new terminal, they must re-export this URL for the program to work correctly; otherwise, it will attempt to use the localhost URL. Therefore, for variables closely related to program execution, it is recommended to save them in a configuration file instead.

I checked `~/.san/providers.json` and found that it only records the models, but does not record the host providing the API functionality.

### Proposed solution

Store runtime-related variables in a fixed file.

### Alternatives considered

_No response_

yanmxa:
Fixes #105 (thanks @lonicerae 🙏).

Base URLs (and the Vertex region/project) were read with raw `os.Getenv`, so a value set in one shell was lost when the terminal closed.

This routes them through `secret.Resolve` — the same env-first store the API keys already use: it checks the environment variable first, then falls back to `~/.san/secrets.json`. So `OLLAMA_BASE_URL` (and the others) can be set once and persist across sessions; env vars still take precedence, so nothing changes for existing setups.

Covers the base URLs for Ollama, DeepSeek, Moonshot, BigModel, Alibaba, MiniMax, plus the Vertex `CLOUD_ML_REGION` / `ANTHROPIC_VERTEX_PROJECT_ID`.

### DIFF
--- a/internal/llm/alibaba/apikey.go
+++ b/internal/llm/alibaba/apikey.go
@@ -2,7 +2,6 @@ package alibaba
 
 import (
 	"context"
-	"os"
 
 	"github.com/openai/openai-go/v3"
 	"github.com/openai/openai-go/v3/option"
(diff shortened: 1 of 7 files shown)


yanmxa:
Please take a look [lonicerae](https://github.com/lonicerae)~

yanmxa:
Thanks again for the clear, well-reasoned report, @lonicerae 🙏 Merging now — your suggestion to persist these in a config file is exactly what this does: env var still takes precedence, then it falls back to `~/.san/secrets.json`.

lonicerae:
LGTM! Thanks @yanmxa ! :thumbsup:

PR #119, opened 2026-06-05
chore: add pre-commit hook and document git hooks setup

zhfeng:
## Summary
- Add `.githooks/pre-commit` that auto-formats staged Go files via `make format` and runs `make format-check` before each commit
- Update `CONTRIBUTING.md` with `core.hooksPath .githooks` setup instructions

## Test plan
- [ ] Clone repo, run `git config core.hooksPath .githooks`
- [ ] Stage a badly formatted `.go` file and verify the hook auto-formats it
- [ ] Verify `make format-check` passes after hook runs

### DIFF
--- a/.githooks/pre-commit
+++ b/.githooks/pre-commit
@@ -39,8 +39,9 @@ echo "$CLEAN_FILES" | xargs goimports -w
 # Re-stage the formatted files
 echo "$CLEAN_FILES" | xargs git add
 
-# Verify formatting
-unformatted=$(echo "$STAGED_GO_FILES" | xargs gofmt -l)
+# Verify the files we just formatted (not skipped ones, whose unstaged
+# changes would falsely block an otherwise-clean commit)
+unformatted=$(echo "$CLEAN_FILES" | xargs gofmt -l)
 if [ -n "$unformatted" ]; then
     echo "These files still aren't formatted — run: make format"
     echo "$unformatted"
--- a/.githooks/pre-commit
+++ b/.githooks/pre-commit
@@ -13,10 +13,11 @@ UNSTAGED=$(git diff --name-only)
 CLEAN_FILES=""
 SKIPPED=""
 for f in $STAGED_GO_FILES; do
-    case "$UNSTAGED" in
-        *"$f"*) SKIPPED="$SKIPPED $f" ;;
-        *) CLEAN_FILES="$CLEAN_FILES $f" ;;
-    esac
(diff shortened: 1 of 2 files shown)


.githooks/pre-commit
```diff
@@ -0,0 +1,24 @@
...
+
+# Get list of staged .go files
+STAGED_GO_FILES=$(git diff --cached --name-only --diff-filter=ACM | grep '\.go$')
+
+if [ -z "$STAGED_GO_FILES" ]; then
+    exit 0
+fi
+
+echo "Running make format..."
+make format
```
yanmxa: **Issue: formats the whole repo, not just staged files.**

`make format` runs `gofmt`/`goimports` over *every* `.go` file in the project (`GOFILES := find . -name '*.go'`, `Makefile:6`), not the staged ones. So every commit reformats and re-checks the entire tree — slow (grows with repo size), and it can silently rewrite unrelated files in the working tree.

Suggestion — scope to staged files only:
```suggestion
echo "$STAGED_GO_FILES" | xargs gofmt -w
echo "$STAGED_GO_FILES" | xargs goimports -w
```
This also makes the whole-repo `make format-check` below redundant — drop it, or replace with a staged-only `gofmt -l` check.

.githooks/pre-commit
```diff
@@ -0,0 +1,24 @@
...
+
+# Re-stage the formatted files
+echo "$STAGED_GO_FILES" | xargs git add
```
yanmxa: **Issue: can silently commit unstaged changes.**

This `git add`s the *whole* file. If a contributor staged only part of a file (`git add -p`), `make format` rewrites the full working-tree copy and this re-stages all of it — pulling the unstaged hunks into the commit without them noticing.

Suggestion — make the hook check-only (no auto-fix, no re-stage), which removes the footgun entirely:
```sh
unformatted=$(echo "$STAGED_GO_FILES" | xargs gofmt -l)
if [ -n "$unformatted" ]; then
  echo "These staged files aren't formatted — run: make format"
  echo "$unformatted"
  exit 1
fi
```
If you'd rather keep auto-formatting, only format files that have no unstaged changes (diff the staged list against `git diff --name-only`) and skip the rest.

yanmxa (dismissed):

yanmxa:
Thanks again for the quick second revision, @zhfeng! 🙌 Both earlier concerns are fully addressed. I pushed a few small follow-up commits for the remaining robustness nits (goimports check, exact path matching, and scoping the verify step). Really appreciate the contribution!

yanmxa (approved):

PR #122, opened 2026-06-05
fix: need to update provider base URL for ollama

lonicerae:
In issue #105, we will be reading the value of OLLAMA_BASE_URL from secrets.json; however, we could potentially update the currently used value back into this file to ensure it can be re-read if the environment variable is missing next time.

### DIFF
--- a/internal/llm/ollama/apikey.go
+++ b/internal/llm/ollama/apikey.go
@@ -36,6 +36,11 @@ func NewAPIKeyClient(ctx context.Context) (llm.Provider, error) {
 		baseURL += "/v1"
 	}
 
+	// Update the secrets file
+	if store := secret.Default(); store != nil {
+		_ = store.Set("OLLAMA_BASE_URL", baseURL)
+	}
+
 	// Ollama ignores the Authorization header; "ollama" is a placeholder.
 	client := openai.NewClient(
 		option.WithAPIKey("ollama"),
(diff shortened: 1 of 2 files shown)


yanmxa:
Thanks for tackling this — the need (#105) is real, but I'd lean **request changes** here. Two issues:

- **Persists the fallback default.** When `OLLAMA_BASE_URL` is unset, `baseURL` defaults to `http://localhost:11434/v1`, and this writes that default into `secrets.json`. Users who configured nothing get a config entry they never set — and stay pinned to a stale default if Ollama's default ever changes.
- **Write churn.** `Set` always calls `save()`, so this rewrites `secrets.json` on every `NewAPIKeyClient` call, even when nothing changed.

Better: only persist a *user-provided* value (not the default), and only when it differs from what's stored.

lonicerae:
Make sense. Will update it.

lonicerae:
@yanmxa , updated a little. Request your review. Thanks!

yanmxa (commented):
Direction is right, and you've already addressed my earlier round — no longer persisting the default, only writing on change. Thanks for the quick turnaround.

Two small notes, neither a blocker:

1. `ResolveOsEnv` just duplicates `os.Getenv`, which already returns `""` when unset — drop the wrapper and call `os.Getenv` directly.

2. #105 covers several runtime vars (Vertex region/project, the other providers' `*_BASE_URL`), so the natural home for persist-on-resolve is `secret.Resolve` itself — then every caller gets it for free instead of each provider inlining the same block. Happy to take that as a follow-up if you'd rather land the Ollama fix first.

PR #129, opened 2026-06-05
fix: set module path to /v2 to match the v2 release tags

yanmxa:
## Problem

The repo has `v2.0.0` / `v2.0.1` tags, but `go.mod` still declared `module github.com/genai-io/san` (no `/v2`). Per Go's [semantic import versioning](https://go.dev/ref/mod#major-version-suffixes), v2+ requires a `/v2` path suffix, so the module proxy won't serve v2 for the bare path. Result:

```
$ curl -s https://proxy.golang.org/github.com/genai-io/san/@latest
{"Version":"v1.19.3", ...}   # @latest resolves to stale v1, not v2
```

So `go install github.com/genai-io/san/cmd/san@latest` installs **v1.19.3**, and pkg.go.dev shows v1.

## Fix

- Bump module path to `github.com/genai-io/san/v2` in `go.mod`.
- Update all internal import paths (309 .go files) to `/v2`.
- Update the `go install` commands in `README.md` / `README.zh.md` / `docs/guides/getting-started.md` / `site/getting-started.html` and the import examples in `docs/cn/*`.
- Update `tools/layercheck` `repoModule` constant so layer checks still match internal packages.
- **GitHub repo URLs are intentionally left unchanged** — only the Go module path is versioned.

763 insertions / 763 deletions — pure path rewrites, no logic changes.

## Verified

`go build ./...`, `go vet ./...`, `gofmt`, and `make lint` (incl. layercheck) all pass; the `san` binary builds and runs.

## Follow-up (after merge)

Cut a new `v2.x.x` tag (e.g. `v2.0.2`) so the proxy serves v2 at `@latest`. Without a fresh tag on the `/v2` module, `@latest` will still resolve to v1.


yanmxa:
Closing in favor of staying on v1 (Strategy A). San is a CLI app, not an imported library, so the /v2 module-path suffix only adds maintenance cost with no benefit. Instead we'll release the current main as v1.20.0 and retire the v2.0.0/v2.0.1 tags. See discussion.

PR #144, opened 2026-06-06
docs(persona): persona system design

yanmxa:
## What

A design doc for **personas**: one on-disk folder that bundles a switchable
**system prompt**, a **skill set**, and a **config overlay**, so the user can
switch the whole bundle in one command, mid-session, without restarting.

It unifies two mechanisms that are independent today (`identity` and `skill`) and
simplifies the system prompt to **four replaceable parts** — `identity` /
`behavior` / `rules` / `environment` — each overridable by a file, falling back to
the built-in default.

Doc lives at `notes/active/persona-system.md`. **Design only — no code changes.**

## Highlights

- **One folder = one persona**: `.san/personas/<name>/{system/, skills/, settings.json}`.
- **First-principles system prompt**: 4 parts (who / how / what-rules / where);
  every prose part replaceable, no special cases. Safety stays enforced in the
  permission layer (defense in depth), not just the prompt.
- **`settings.json` overlay** reuses `setting.Data` + `merger.go` as the highest
  file-level layer; `skills`/`disabledTools` are per-key overrides, permission
  `deny` is add-only (a persona can tighten but never loosen).
- **Two-tier switch**: prompt + skills apply immediately (hot-patch); tools +
  permissions apply on the next agent rebuild.
- **Additive**: reuses existing hot-swap, reminder, and merge machinery; does not
  touch the agent run loop, LLM interface, or session persistence.

## Contents

§0 Motivation · §1 Concept · §2 First principles (system prompt) · §3 On-disk
layout · §4 settings.json overlay · §5 Switch flow · §6 Relationship to
identity & skills · §7 Current architecture · §8 Decisions (locked) · §9 Phasing
· §10 Open questions

## Status

Seeking review on the **design** before implementation. §9 lists the phased
rollout (7 steps, each independently shippable; step 1 — simplify the slots —
carries no behavior change).

### DIFF
--- a/notes/active/persona-system.md
+++ b/notes/active/persona-system.md
@@ -1,0 +1,392 @@
+# Personas — switchable system prompt + skills + config
+
+A **persona** is a single on-disk folder that bundles everything that makes San
+behave a certain way — its system prompt, its skills, and its config overrides —
+so the user can switch the whole bundle in one command, mid-session, without
+restarting.
+
+This unifies two concepts that are independent today (`identity` and `skill`),
+and simplifies the system prompt down to a few replaceable parts.
+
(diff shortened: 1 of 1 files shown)


ldpliu:
Two questions on the design:

1. **Default persona config.** Should the default persona's configuration be exposed to users? If exposed, could a user manually modify it? Or would the system revert/rollback any changes to prevent drift from the built-in defaults?

2. **MCP config in persona.** Should MCP server configuration (`mcp.json`) be part of the persona bundle? It feels like a natural fit alongside skills and tools — a persona might want to bring its own MCP servers (e.g., a "research" persona with a papers-search MCP server). But it also adds scope. Curious what others think.

yanmxa:
I'm intent to keep the default invisible. For the first version we can start with the core identity and skills to keep it simply

ldpliu:
> I'm intent to keep the default invisible. For the first version we can start with the core identity and skills to keep it simply

That's fine

PR #150, opened 2026-06-07
fix(mimo): add provider to UI list and use secret.Resolve for base URL

zhfeng:
## What & why

The MIMO provider was registered at startup but missing from the hardcoded `providerOrder` and `providerDisplayNames` lists in the provider UI, so it never appeared in the `/model` selector.

Also switches `MIMO_BASE_URL` lookup from `os.Getenv` to `secret.Resolve` for consistency with other providers (supports both env vars and `secrets.json`).

Closes #

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor / cleanup
- [ ] Docs
- [ ] Other:

## Testing

Verified MIMO appears in the provider list and `secret.Resolve` correctly reads `MIMO_BASE_URL` from environment.

```bash
```

## Checklist

- [x] Follows [Conventional Commits](https://www.conventionalcommits.org/) (`fix:`)
- [x] Tests pass locally

### DIFF
--- a/internal/app/input/on_provider.go
+++ b/internal/app/input/on_provider.go
@@ -743,6 +743,7 @@ var providerOrder = []llm.Name{
 	llm.Alibaba,
 	llm.BigModel,
 	llm.Ollama,
+	llm.Mimo,
 }
 
 // providerDisplayNames maps provider to human-readable name.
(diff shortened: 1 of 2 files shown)


yanmxa (approved):

zhfeng:
@yanmxa - it looks like enhancement to avoid the hardcode of the providerOrder and providerDisplayNames, is it possible to be dynamically from the provider registry?

zhfeng:
I open the following PR
- https://github.com/genai-io/san/pull/151

Issue #152, opened 2026-06-07
Cannot update API key for providers


zhfeng:
## Problem

The providers list is read-only and there is no way to update the API key once it has been set. If a user needs to change or rotate their API key, there is no UI to do so.

yanmxa:
This should be fixed by https://github.com/genai-io/san/pull/154. Thank you @laisongls, for addressing it

PR #154, opened 2026-06-07
fix: enable-editing-for-api-keys

laisongls:
Fix issue https://github.com/genai-io/san/issues/152

🐛 Bug Confirmation
The issue was in the provider selector UI logic. Once a provider was connected (StatusConnected), selecting its auth method would only trigger a refresh (refreshAuthMethod) instead of showing the API key input field again. This made it impossible to edit API keys after the initial setup.

🔍 Root Cause
In the selectAuthMethod function, connected auth methods were hardcoded to only refresh the connection rather than allowing API key re-entry.

💡 Why This Fixes It
Preserves existing functionality: For connected providers with valid env vars, it will still work as before.
Enables API key editing: For connected providers where users want to change credentials, it now shows the API key input field.

🛠️ Implementation Details
on_provider.go - Added handleCredentialEdit functionality:
Added 'e' key handling in HandleKeypress to trigger the edit action.
Fixed logic in the tea.KeyRunes case to ensure the 'e' key correctly triggers editing (instead of being appended to the search query) when the search bar is empty.
Implemented handleCredentialEdit() to handle editing for both providers and auth methods.
Implemented handleCredentialEditForProvider() to handle providers with single or multiple auth methods.
Implemented handleCredentialEditForAuthMethod() to handle auth method editing.

✨ User Experience
Users can now:
Navigate to the Providers tab in the /model interface.
Select a connected provider.
Press the e key to enter API Key edit mode.
Input the new API Key.
Press Enter to submit. The system will update the environment variables and reconnect the provider.

✅ Testing
All tests passed.

### DIFF
--- a/internal/app/input/on_provider.go
+++ b/internal/app/input/on_provider.go
@@ -479,6 +479,9 @@ func (s *ProviderSelector) HandleKeypress(key tea.KeyMsg) tea.Cmd {
 	case tea.KeyRunes:
 		s.appendModelSearch(string(key.Runes))
 		return nil
+
+	case tea.KeyCtrlE:
+		return s.handleCredentialEdit()
 	}
 
 	// Vim navigation (only when search query is empty)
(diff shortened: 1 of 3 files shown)


yanmxa:
Thanks a lot for tackling this, @laisongls — really appreciate the contribution! 🙏 The fix is clean and well-tested (nice coverage on the single/multiple auth-method paths and the env-var edge cases), and it solves #152 nicely.

One small, non-blocking note on the `e` shortcut: it's intercepted whenever the search box is empty, regardless of the active tab. Since both tabs support type-to-search, on the **Models** tab this swallows `e` as the first search character — you can't start a model search with `e` (it's consumed instead of typed).

A clean way around this is to use a capital `E` for the edit shortcut and scope it to the Providers tab. Search is case-insensitive (everything is lowercased before matching), so an uppercase letter won't collide with typing a query:

```go
case tea.KeyRunes:
    if s.activeTab == providerTabProviders && string(key.Runes) == "E" {
        return s.handleCredentialEdit()
    }
    s.appendModelSearch(string(key.Runes))
    return nil
```

That also lets you drop the fall-through into the vim-navigation block. Worth updating the hint to match (`on_provider_view.go`): `"e edit"` → `"E edit"`.

Totally optional, though — happy to merge as-is. Thanks again for the great work!

zhfeng:
I'd like to use `e` to enter the edit mode. For search models, can we trigger it by `/`?

yanmxa:
@zhfeng After discussing with @laisongls,we both agree combining ctrl with e might be a better choice🤔

yanmxa:
> I'd like to use `e` to enter the edit mode. For search models, can we trigger it by `/`? 

Search with slash will break the default behavior for all the search panels 😅

laisongls:
Guys, use Ctrl+E is the best practice, it will not be affected by the search function.
- Add handleCredentialEdit function to allow editing API keys for connected providers
- Add Ctrl+E keyboard shortcut to trigger credential edit mode on Providers tab  
- Implement handleCredentialEditForProvider and handleCredentialEditForAuthMethod helper functions
- Update UI hints to show Ctrl+E edit shortcut on Providers tab
- Add comprehensive tests for credential editing functionality

internal/app/input/on_provider.go
```diff
@@ -743,7 +818,6 @@ var providerOrder = []llm.Name{
 	llm.Alibaba,
 	llm.BigModel,
 	llm.Ollama,
-	llm.Mimo,
```
yanmxa, line 746: Is this the intent?

yanmxa: @laisongls Please rebase your changes on the current main branch~

yanmxa (approved):

yanmxa:
Hi @laisongls , Please squash the commits into one with a sign-off, It's almost landing!

zhfeng:
Nice work - I checked  locally and it works as expected. Aslo is it possible to add another key to remove the API_KEY as well? e.g. `Ctrl-D` for removing the API_KEY

zhfeng:
I had worked on removing the key with a comfirm prompt. I will open a follow PR after this one is merged.

yanmxa:
@zhfeng Thank you for working on this and for planning the following PR. The confirmation prompt for API removal will be a nice addition!

yanmxa (approved):

yanmxa:
The changes look good, and I appreciate your effort！

zhfeng:
@yanmxa - The following PR for removing AP_KEY
- https://github.com/genai-io/san/pull/158

PR #158, opened 2026-06-07
feat(provider): add Ctrl+D to remove API key with confirmation

zhfeng:
## Summary
- Add Ctrl+D shortcut to remove stored API keys from the provider selector
- Shows an inline confirmation prompt (y/N) before removing
- On confirm: removes key from secret store and env, disconnects provider, clears cached models, and resets current model if it belonged to the removed provider
- Adds `Disconnect`, `RemoveCachedModels`, and `ClearCurrentModel` methods to `llm.Store`
- Adds `Delete` method to `secret.Store`

## Test plan
- [x] Ctrl+D on single-auth provider shows confirmation, 'y' removes key
- [x] Ctrl+D on specific auth method row removes only that credential
- [x] Any key other than 'y' cancels the removal
- [x] Models and connection are cleared from store after removal
- [x] Non-providers tab ignores Ctrl+D

### DIFF
--- a/internal/app/input/on_provider.go
+++ b/internal/app/input/on_provider.go
@@ -203,6 +203,11 @@ type ProviderSelector struct {
 	apiKeyProviderIdx int // index into allProviders
 	apiKeyAuthIdx     int // index into that provider's AuthMethods
 
+	// Inline confirm-remove prompt
+	confirmRemoveActive  bool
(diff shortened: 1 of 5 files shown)


yanmxa:
Thanks @zhfeng, Merged it！

PR #148, opened 2026-06-06
feat(release): add Windows builds to release target

zhfeng:
## What & why

Add Windows (amd64 + arm64) cross-compilation to the `make release` target so Windows users can download pre-built binaries without needing a Go toolchain.

Windows binaries are packaged as `.zip` archives (the expected format on Windows), alongside the existing `.tar.gz` tarballs for macOS and Linux.

Closes #48

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Refactor / cleanup
- [ ] Docs
- [ ] Other:

## Testing

Cross-compilation verified on macOS:

```bash
GOOS=windows GOARCH=amd64 go build -ldflags "-s -w" -o san.exe ./cmd/san
file san.exe
# PE32+ executable (console) x86-64, for MS Windows
```

## Checklist

- [x] Follows [Conventional Commits](https://www.conventionalcommits.org/) (`feat:`)
- [x] All commits signed off (`git commit -s`) — certifies the [DCO](https://developercertificate.org)
- [x] Tests pass locally
- [ ] Docs updated (if behavior or config changed)
- [x] Read and follow the [Code of Conduct](CODE_OF_CONDUCT.md)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #159, opened 2026-06-07
feat(install): add Windows PowerShell installer and publish zip artifacts

yanmxa:
## What & why

Follow-up to #148, which added Windows `.zip` builds to `make release` but left two gaps that meant Windows users still couldn't get a binary:

- **Publish the zips.** The release workflow only uploaded `bin/san_*.tar.gz`, so the newly-built Windows `.zip` archives were dropped at publish time. Added `bin/san_*.zip` to the upload globs in `release.yml`.
- **Install script.** Added `install.ps1`, a PowerShell installer mirroring `install.sh` (`install` / `upgrade` / `uninstall`): detects arch, downloads the Windows zip, installs `san.exe` to `%LOCALAPPDATA%\san\bin`, and adds it to the user PATH.
- **Docs.** Documented the PowerShell one-liner under the README installation methods.

## Testing

PowerShell installer reviewed for PS 5.1 compatibility (`Invoke-RestMethod`, `Expand-Archive`, WOW64 arch detection). End-to-end download path can only be exercised once a release ships Windows assets — the next tagged release will be the first to include them.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -53,5 +53,6 @@ jobs:
         with:
           files: |
             bin/san_*.tar.gz
+            bin/san_*.zip
             CHANGELOG.md
           body_path: release-notes.md
--- a/README.md
+++ b/README.md
@@ -63,6 +63,18 @@ curl -fsSL https://raw.githubusercontent.com/genai-io/san/main/install.sh | bash
 <details>
 <summary><b>Other methods</b></summary>
 
+**Windows (PowerShell)**
+
+```powershell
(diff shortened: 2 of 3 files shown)


zhfeng (approved):
LGTM

lonicerae:
LGTM. BTW, please also check if the confirmation should be added as well. https://github.com/genai-io/san/pull/162/changes/0fbe803c1a495f1c54867286afc475d43a6c00f8#diff-043df5bdbf6639d7a77e1d44c5226fd7371e5259a1e4df3a0dd5d64c30dca44fR135

lonicerae (approved):

PR #161, opened 2026-06-08
fix: remove hardcoded model fallback when no model selected

skeeey:
When a provider credential is removed via Ctrl+D in the Providers tab, the store's CurrentModel was cleared but the in-memory env.CurrentModel and llm.Conn.currentModel stayed stale, causing the footer and header to display the old (now-orphaned) model name.

This also fixes the hardcoded 'claude-sonnet-4-20250514' fallback in GetModelID() — it now returns empty when no model is configured, and the display layer shows 'no model selected'.

**Changes:**
- `env.GetModelID()`: returns `""` instead of the hardcoded model when `CurrentModel` is nil
- `env.GetModelDisplayName()`: returns `"no model selected"` when no model is active
- Clear `llm.Default().SetCurrentModel(nil)` and `llm.Default().SetProvider(nil)` on credential removal
- Sync `env.CurrentModel` and `env.LLMProvider` from `llm.Default()` on popup dismiss (`DismissedMsg`)
- Switch view layer to use `GetModelDisplayName()` so footer and conversation header consistently show the right message
- Sync `llm.Default().SetCurrentModel()` on model selection so all three state layers (store, llm.Conn, env) stay consistent

### DIFF
--- a/internal/app/update.go
+++ b/internal/app/update.go
@@ -81,8 +81,19 @@ func (m *model) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
 	case input.PromptSuggestionMsg:
 		input.HandlePromptSuggestion(&m.userInput, m.conv.Stream.Active, m.userInput.Textarea.Value(), msg)
 		return m, nil
-	case kit.DismissedMsg, input.ToolToggleMsg:
+	case kit.DismissedMsg:
+		// Sync env state with llm.Default. Credential removal (Ctrl+D in
+		// the Providers tab) clears the store's CurrentModel and updates
+		// llm.Default — but the per-turn env fields would stay stale until
+		// the next explicit model selection without this refresh.
+		prevModel := m.env.CurrentModel
+		m.env.CurrentModel = m.services.LLM.CurrentModel()
+		m.env.LLMProvider = m.services.LLM.Provider()
+		if prevModel != nil && m.env.CurrentModel == nil {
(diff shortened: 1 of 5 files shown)


yanmxa:
Thanks for contributing! 

@skeeey It looks good to me. Is that ok, I'm going to merge it now?

skeeey:
@yanmxa, I think we have a whole fix for this part now, please take a look again, thanks


/unhold

internal/app/update.go
```diff
@@ -81,8 +81,19 @@ func (m *model) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
...
+		// llm.Default — but the per-turn env fields would stay stale until
+		// the next explicit model selection without this refresh.
+		prevModel := m.env.CurrentModel
+		m.env.CurrentModel = m.services.LLM.CurrentModel()
+		m.env.LLMProvider = m.services.LLM.Provider()
+		if prevModel != nil && m.env.CurrentModel == nil {
+			return m, m.overlayDeps().PrintWelcome(m.env.GetModelDisplayName())
+		}
 		return m, nil
+	case input.ToolToggleMsg:
```
yanmxa, line 96: Empty case body with no `return` — `ToolToggleMsg` now falls past the switch into `routeToSubModel`/`updateTextarea` instead of returning `m, nil`. Harmless today, but unintended. Suggest:

```go
case input.ToolToggleMsg:
    return m, nil
```

PR #168, opened 2026-06-08
fix(mimo): populate InputTokens from CacheReadInputTokens

zhfeng:
## What & why

Mimo reports all input tokens as `CacheReadInputTokens`, leaving `InputTokens` at 0. This breaks downstream consumers (logging, cost tracking) that read `InputTokens`. The fix intercepts the final `Done` chunk and copies `CacheReadInputTokens` into `InputTokens` when the latter is zero.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor / cleanup
- [ ] Docs
- [ ] Other:

## Testing

Verified by observing token counts in the CLI output — `InputTokens` now correctly reflects the actual input token count instead of showing 0.

## Checklist

- [x] Tests pass locally

### DIFF
--- a/internal/llm/mimo/client.go
+++ b/internal/llm/mimo/client.go
@@ -35,8 +35,22 @@ func NewClient(client anthropicsdk.Client, name string) *Client {
 func (c *Client) Name() string { return c.inner.Name() }
 
 // Stream delegates to the anthropic provider for full API compatibility.
+// Mimo reports all input tokens as CacheReadInputTokens; fix up InputTokens
+// so downstream consumers (logging, cost tracking) see the correct count.
 func (c *Client) Stream(ctx context.Context, opts llm.CompletionOptions) <-chan llm.StreamChunk {
-	return c.inner.Stream(ctx, opts)
+	in := c.inner.Stream(ctx, opts)
+	out := make(chan llm.StreamChunk)
+	go func() {
+		defer close(out)
+		for chunk := range in {
+			if chunk.Type == llm.ChunkTypeDone && chunk.Response != nil &&
+				chunk.Response.Usage.InputTokens == 0 && chunk.Response.Usage.CacheReadInputTokens > 0 {
+				chunk.Response.Usage.InputTokens = chunk.Response.Usage.CacheReadInputTokens
+			}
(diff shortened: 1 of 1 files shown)


yanmxa:
Root cause is upstream — Mimo mislabels all input as `cache_read_input_tokens`, so ideally this is fixed in their API (worth filing an issue with them).

That said, a guarded shim here is reasonable (this package already adapts other Mimo quirks, and the `InputTokens == 0` check makes it a no-op once they fix it). Two things if we keep it:

- Also handle `CacheCreationInputTokens` — on cache-miss turns the tokens land there, so `InputTokens` would still be 0.
- Move, don't copy — `EstimateCost` sums both `InputTokens` and `CacheReadInputTokens`, so copying double-counts those tokens. Zero out the cache field after assigning.

yanmxa:
Closing in favor of fixing this upstream — it's a Mimo API bug (input mislabeled as cache_read_input_tokens), not really ours to patch per-provider. Will file an issue with them. We can revisit a corrected shim here if we need it before they fix it.

zhfeng:
Yeah, I see it is an upstream issue. But unfornatelly I can not find where to open an issue with them. I only find the docs like https://platform.xiaomimimo.com/docs/en-US/api/chat/anthropic-api

zhfeng:
@yanmxa can you add a `hold` label before we find a way to open an upstream issue? Thanks!

wangbin77:
This is not a bug. In anthropic api (https://platform.claude.com/docs/en/build-with-claude/prompt-caching)
total_input_tokens = cache_read_input_tokens + cache_creation_input_tokens + input_tokens

so input_tokens doesn't include cache_read_input_tokens.

yanmxa:
> This is not a bug. In anthropic api (https://platform.claude.com/docs/en/build-with-claude/prompt-caching) total_input_tokens = cache_read_input_tokens + cache_creation_input_tokens + input_tokens
> 
> so input_tokens doesn't include cache_read_input_tokens.

Yes, can you post the upstream issue link? And is it okay to close this one?

yanmxa:
Closing this for now — will reopen if needed.

PR #170, opened 2026-06-09
fix(tui): commit streaming output to scrollback at newline boundaries

ldpliu:
## User-Facing Change

Previously, streaming output was confined to a small fixed viewport. Once the output exceeded the viewport size, the earlier lines were clipped and completely inaccessible — users had to wait until the full response finished before they could scroll back to read anything that had scrolled off screen.

Now, like the Claude Code CLI, the entire terminal scrolls naturally as output streams in. Completed lines are progressively flushed to native terminal scrollback, so users can freely use the mouse wheel (or trackpad scroll) to review earlier content at any point during streaming — no need to wait for the response to complete.

## Problem

PR #77 simplified the scrolling architecture by removing the in-app scroll viewport and mouse capture, relying on native terminal scrollback. The `tailLines()` cap shows the latest lines of the streaming output in View(), but drops earlier lines — with no way for the user to scroll up to them.

## Root Cause

During streaming, the active (uncommitted) assistant message stays entirely in View(). `tailLines()` truncates it to the available terminal height, showing only the last N lines. The truncated top portion is neither in View() nor in scrollback — the mouse wheel cannot reach it.

Additionally, View() re-renders every frame during streaming. The cursor movement from bubbletea's re-rendering causes the terminal to auto-scroll, making it impossible to maintain a scrolled-up position even for content that IS in scrollback.

## Solution

Restore incremental scrollback commits with proper deduplication:

1. **`commitStreamTail`**: flush completed lines of both thinking and content text to native terminal scrollback at newline boundaries during streaming. Thinking text is committed once (when content first appears). At stream end, `CommittedCount` is advanced to skip `CommitMessages` — preventing duplicate output.

2. **`ScrollbackLen` in `RenderContext`**: tracks bytes of the last assistant message already committed to scrollback. The View() renderer uses this to only draw the **uncommitted tail**, keeping the live area small so cursor movement does not trigger terminal auto-scroll.

3. **Thinking clearing**: when `ScrollbackLen > 0` (content is streaming), the thinking text is cleared from View() rendering — the thinking phase is over and re-rendering it clutters the live view.

## Changes

| File | Change |
|------|--------|
| `conv/conversation.go` | Add `ScrollbackLen` and `ThinkingCommitted` to `StreamState` |
| `conv/update.go` | Add `commitStreamTail` function, call it from `applyChunk` |
| `conv/view.go` | Add `ScrollbackLen` to `RenderContext`, trim content and clear thinking in `RenderMessageAt` |
| `app/view.go` | Pass `ScrollbackLen` through `messageRenderParams` |

## Behavior

- **Thinking phase**: thinking text shows briefly in View(), then committed to scrollback once, hidden from View()
- **Content streaming**: completed lines go to scrollback (mouse-wheel accessible), only the incomplete last line stays in View() (small, stable)
- **Stream end**: remaining tail flushed to scrollback, no duplicate output

### DIFF
--- a/internal/app/conv/conversation.go
+++ b/internal/app/conv/conversation.go
@@ -5,13 +5,17 @@ import (
 )
 
 type StreamState struct {
-	Active       bool
-	BuildingTool string
+	Active            bool
+	BuildingTool      string
+	ScrollbackLen     int  // bytes of last assistant message already committed to scrollback
+	ThinkingCommitted bool // whether thinking text has been committed to scrollback
 }
 
 func (s *StreamState) Stop() {
 	s.Active = false
 	s.BuildingTool = ""
+	s.ScrollbackLen = 0
+	s.ThinkingCommitted = false
 }
 
 type ConversationModel struct {
--- a/internal/app/conv/markdown.go
+++ b/internal/app/conv/markdown.go
@@ -157,6 +157,25 @@ func isTableLine(line string) bool {
 	return strings.HasPrefix(trimmed, "|") && strings.Contains(trimmed[1:], "|")
 }
 
+// isAllTableLines reports whether every non-empty line in s looks like a
+// markdown table row.  Returns false when s is empty or contains only
+// whitespace.
+func isAllTableLines(s string) bool {
+	lines := strings.Split(s, "\n")
+	found := false
+	for _, line := range lines {
+		trimmed := strings.TrimSpace(line)
+		if trimmed == "" {
+			continue
+		}
+		found = true
+		if !isTableLine(trimmed) {
+			return false
+		}
+	}
+	return found
+}
+
 // findTableEnd finds the end index (exclusive) of consecutive table lines.
 func findTableEnd(lines []string, start int) int {
 	i := start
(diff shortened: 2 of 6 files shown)


ldpliu:
https://github.com/user-attachments/assets/35b55902-8de0-4a42-b0ca-d551b38172da

ldpliu:
## Before/After Comparison

The video below demonstrates the fix in action — **main branch (left)** vs **fixed code (right)**:

- **Left (main)**: During streaming, once output exceeds the terminal window height, earlier lines are completely inaccessible. Users are forced to wait until the entire response finishes before they can scroll back to see anything that scrolled off screen.
- **Right (fixed)**: Completed lines are progressively flushed to native terminal scrollback. Users can freely scroll up at any point during streaming to review earlier content — no need to wait for the response to complete.

yanmxa:
It seems the table is not correctly rendered

image

ldpliu:
image

yanmxa:
As we discussed yesterday — this works for now, but it's a bit of a hack. We'll come back and fix it properly once we find a cleaner approach.

PR #187, opened 2026-06-12
feat(hook): run PreToolUse before main tools

zhfeng:
## Summary
- Run PreToolUse hooks in the main-session tool execution path
- Apply hook-updated tool input before permission checks and execution
- Honor PreToolUse permission allow/ask decisions and cover them with tests

## Test plan
- [x] go test ./internal/tool ./internal/tool/perm ./internal/agent ./internal/app ./internal/setting ./internal/hook ./tests/integration/hooks

### DIFF
--- a/internal/agent/build.go
+++ b/internal/agent/build.go
@@ -7,6 +7,7 @@ import (
 
 	"github.com/genai-io/san/internal/core"
 	"github.com/genai-io/san/internal/core/system"
+	"github.com/genai-io/san/internal/hook"
 	"github.com/genai-io/san/internal/llm"
 	"github.com/genai-io/san/internal/tool"
 )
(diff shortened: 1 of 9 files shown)


yanmxa (commented):
Direction looks right (hooks before permission, updated input applied before the check). One issue I'd consider blocking:

**The `perm.WithDecision` context override leaks into nested tool calls.** When the wrapped tool is the Agent tool, the ctx (with a `Permit` override) flows into the subagent run, and the subagent's own `WithPermission` tools read `perm.DecisionFromContext` first — so one hook "allow" silently bypasses permission checks for every tool the subagent executes.

Suggest handling the hook outcome with local control flow in the wrapper (block / allow / force-prompt / normal check) instead of passing decisions through context — e.g. give the bridge an explicit force-prompt entry that can also carry the hook's reason into the prompt.

yanmxa:
Rebased onto latest main and pushed a small follow-up commit (c40675e) on top of your work, addressing a few permission edge cases: skip the decider when a hook forces a prompt (it was discarded but still emitted an audit record), let a hook "ask" reach the user even if another hook answered "allow", and use a consistent "blocked:" error prefix. Your commit and sign-off are unchanged. Feel free to adjust if you'd rather handle any of these differently.

zhfeng:
Thanks @yanmxa - LGTM!

PR #190, opened 2026-06-12
feat(hook): run PreToolUse hooks before main-session tools

yanmxa:
Run PreToolUse hooks ahead of permission checks for main-session tool execution. Alternative to #187 (thanks @zhfeng for the groundwork) that addresses the review feedback there: hook decisions are handled with local control flow in a single wrapper instead of being passed through context, so a hook "allow" on an Agent call cannot leak into the subagent's permission checks.

- One gate wrapper runs hook → block / input rewrite / allow / force-ask → permission check; `WithPermission` is unchanged for subagents
- "ask" goes through an explicit `PermissionBridge.ForcePromptFunc` that shows the hook's reason in the prompt; "ask" wins over "allow" when both are returned
- Regression test asserts a hook allow on an outer call does not bypass nested tool calls' permission checks

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,37 @@
 All notable changes to San are documented here.
 The format follows [Keep a Changelog](https://keepachangelog.com/en/1.1.0/), and this project adheres to [Semantic Versioning](https://semver.org/).
 
+## [v1.20.1] - 2026-06-11
+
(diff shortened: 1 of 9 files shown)


yanmxa:
Closing — let's address this in #187 instead. Feel free to borrow anything from this branch.

PR #196, opened 2026-06-13
feat(cli): add --persona flag to activate a persona on startup

ldpliu:
## Summary

Adds a `--persona` CLI flag so users can start san with a specific persona:

```bash
san --persona backend-dev                    # activate persona on startup
san --persona backend-dev "write a handler"  # with initial prompt
san --persona nonexistent                    # clear error: "not found; available: ..."
```

## Changes

- **`cmd/san/main.go`** — new `--persona` flag, passed to `RunOptions`
- **`internal/setting/run_options.go`** — new `Persona` field
- **`internal/app/run.go`** — early validation before TTY setup + `validatePersona()` helper
- **`internal/app/model_lifecycle.go`** — persona validation and activation in `applyRunOptions`
- **18 integration tests** in `tests/integration/persona/` covering registry loading, parsing, scope priority, and CLI flag behavior

## Behavior

- Valid persona → activates it (same as `/persona <name>` after startup)
- Nonexistent persona → clear error listing available personas
- `--help` includes the new flag

## Test plan

- [x] All 18 persona integration tests pass
- [x] All existing integration tests pass (cli, session, skill, plugin, etc.)
- [x] `go build ./...` compiles cleanly

### DIFF
--- a/internal/app/run.go
+++ b/internal/app/run.go
@@ -5,6 +5,7 @@ import (
 	"context"
 	"fmt"
 	"os"
+	"strings"
 
 	tea "github.com/charmbracelet/bubbletea"
 
@@ -13,6 +14,7 @@ import (
 	"github.com/genai-io/san/internal/core"
 	"github.com/genai-io/san/internal/hook"
 	"github.com/genai-io/san/internal/llm"
+	"github.com/genai-io/san/internal/persona"
 	"github.com/genai-io/san/internal/setting"
 	"github.com/genai-io/san/internal/tool"
 )
@@ -21,6 +23,12 @@ import (
 func Run(opts setting.RunOptions) error {
 	if opts.Print != "" {
 		return runPrint(opts.Print)
+	}
+
+	if opts.Persona != "" {
+		if err := validatePersona(opts.Persona); err != nil {
+			return err
+		}
 	}
 
 	if userQuit, err := kit.ResolveTheme(setting.LoadTheme(), setting.SaveTheme); userQuit || err != nil {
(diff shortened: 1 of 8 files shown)


yanmxa (commented):
Two points left as inline comments on `internal/app/run.go`.

internal/app/run.go
```diff
@@ -23,6 +25,12 @@ func Run(opts setting.RunOptions) error {
 		return runPrint(opts.Print)
 	}
 
+	if opts.Persona != "" {
```
yanmxa, line 23: Only reached in interactive mode — the `opts.Print != ""` branch above returns first, so `san --persona X -p "..."` never validates *or* applies the persona. `san --persona bogus -p hi` silently succeeds, unlike the interactive path. Move this check above the print branch, or reject `--persona` when `-p` is set.

internal/app/run.go
```diff
@@ -186,3 +194,28 @@ func runPrint(userMessage string) error {
 
 	return nil
 }
+
+// validatePersona checks that the named persona exists on disk before the TUI
+// starts. It initializes the persona registry early (idempotent) and returns a
+// descriptive error listing available personas when the name is not found.
+func validatePersona(name string) error {
```
yanmxa, line 199: Minor: this duplicates the validation block in `applyRunOptions` (`model_lifecycle.go`) almost verbatim — same available-names collection and the same two error strings. Worth extracting one shared helper so the messages stay in sync.

internal/app/run.go
ldpliu: Done: moved `validatePersona` above the `Print` branch in `Run()` so persona validation runs before either mode. Added `TestCLI_NonexistentPersonaWithPrint` to cover `san --persona bogus -p "hello"` — it now fails with "not found" instead of silently proceeding.

internal/app/run.go
ldpliu: Done: extracted `Registry.Validate(name string) error` in the persona package. Both `validatePersona` (run.go) and `applyRunOptions` (model_lifecycle.go) now delegate to it, and `"strings"` import is removed from both files.

yanmxa:
Both review points are resolved and CI is green (lint, test, DCO all pass). Thanks @ldpliu for the quick turnaround — extracting `Registry.Validate` and moving the check above the print branch is exactly right. Merging.

PR #214, opened 2026-06-14
refactor(app): extract modal selectors into internal/app/selector

yanmxa:
Split the 12 modal selectors out of `internal/app/input` into a new `internal/app/selector` package.

`input` had grown to ~13k LOC / 39 files — over half the app layer — by holding selectors (model picker, plugin/MCP managers, provider wizard, memory editor, config panel, approval flow, …) unrelated to text input. After the split, `input` owns only true input concerns (textarea, history, suggestions, queue, images, slash commands).

- `selector.State` aggregates the 12 selectors; `input.Model` embeds it as the `Selectors` field
- `selector.Runtime` (was the in-`input` `OverlayDeps`) carries per-update callbacks, mirroring `conv.Runtime`
- `selector.Registries` carries construction-time registries
- the lone `input.Model` back-reference (textarea prefill in the MCP handler) becomes a `SetInput` callback, so `selector` never imports `input`/`app`

Pure move, no behavior change. build / vet / gofmt / test / layercheck all pass.


yanmxa:
Closing — after review we've decided to keep the modal selectors flat in the input package rather than extracting a separate selector package. The separate package forced a State aggregate + embedding to dodge an import cycle, which added more indirection than the split was worth. The input branch is preserved if we revisit.

PR #216, opened 2026-06-14
feat(tui): real-time context-usage status bar

wangke19:
## Summary

Add a permanent, single-line context-usage status bar that shows how full the model's context window is — updated after every assistant turn.

- **10-cell bar** with percentage label: `[██████░░░░] 71%`
- **4-tier color escalation**: green (healthy) → yellow (watch) → red (pressure) → red+bold (imminent compression)
- **Compressions badge**: `🗜️ N` with escalating color
- **Responsive segment allocator**: tail-budget allocator drops low-priority segments on narrow terminals, never wraps
- **Unknown-limit placeholder**: renders `ctx X/-- [----------] --` (dim) when a model's context limit can't be resolved, instead of silently hiding — makes the gap visible and actionable via `/tokenlimit`

### Provider limit fixes

The status bar needs `InputLimit > 0` to render. Several providers' APIs don't include context-window metadata, so the bar was blank for their models. Fixed per-provider:

| Provider | Issue | Fix |
|----------|-------|-----|
| **SenseNova** | Anthropic-compat endpoint returned 0 tokens in every SSE event | Switched to OpenAI-compat endpoint (`stream_options.include_usage`); added `deepseek-v4-flash` to catalog |
| **BigModel (GLM)** | `/v1/models` omits `context_length` | `staticInputLimit()` fallback: glm-5.2 → 1M, glm-* → 256K |
| **Moonshot** | API returns `max_output_tokens` but not input limit | `staticInputLimit()` parses context from model ID suffix (8k/32k/128k/k2) |
| **OpenAI** | `openAIModelInfo` had no limits at all | `openAILimits()` lookup: gpt-6/5.x/codex → 400K, o-series → 200K, gpt-4.1 → 1M |
| **Anthropic** | All 12 catalog entries lacked `OutputTokenLimit` | Added output limits (Opus → 32K, Sonnet 4.x → 64K, Haiku → 8–16K) |
| **Volcengine** | Env-var model had no limits | `staticLimits()` matches Doubao ID suffixes (256k/128k/32k) |

### Catalog infrastructure

- **`catalog_check_test.go`** — guardrail test that fails if any model in a static catalog has `InputTokenLimit == 0`, preventing silent status-bar breakage when new models are added
- **`CATALOGS.md`** — documents the four catalog patterns (flat, with-pricing, API+fallback, prefix-matching) so new providers follow the right one

### Color tiers (PRD §7.2)

| Range | Tier | Theme token | Meaning |
|-------|------|-------------|---------|
| pct == null | tierNone | Muted | No data yet |
| [0, 50] | tierGood | Success | Healthy |
| (50, 80] | tierWarn | Warning | Watch |
| (80, 95) | tierBad | Error | Pressure |
| [95, 100+] | tierCritical | Error + Bold | Imminent compression |

### Files changed

| Subsystem | Files | What |
|-----------|-------|------|
| Rendering | `conv/status_bar.go`, `conv/status_bar_test.go`, `conv/message.go`, `conv/message_test.go` | Bar, classifier, allocator, badge; wired into `RenderModeStatus`; placeholder when unknown |
| State | `env.go`, `env_test.go`, `model_compact.go`, `view.go` | Compressions counter, token usage threading |
| SenseNova | `sensenova/apikey.go`, `sensenova/catalog.go`, `sensenova/client.go`, `sensenova/client_test.go` | OpenAI SDK switch for real usage; deepseek-v4-flash added |
| Provider limits | `bigmodel/catalog.go`, `moonshot/catalog.go`, `moonshot/client.go`, `openai/catalog.go`, `anthropic/catalog.go`, `volcengine/catalog.go` | Static limit fallbacks + output limit coverage |
| Catalog infra | `llm/catalog_check_test.go`, `llm/CATALOGS.md` | Guardrail test + pattern docs |
| Docs | `notes/completed/status-bar-context-usage.md`, `notes/completed/status-bar-context-usage-plan.md` | Spec + plan moved to completed |

### Test plan

- [x] `TestClassifyContextTier_Boundaries` — 50/80/95 thresholds with off-by-one preservation
- [x] `TestRenderContextBar_FillLevels` — 0%, 50%, 95%, 100%, >100% clamping
- [x] `TestRenderContextBar_NoLimit` — renders `[----------] --` when denominator unknown
- [x] `TestRenderContextBar_NoNegativePercent` — never renders negative
- [x] `TestRenderContextLabel` — compact humanized numbers (`12.4k`, `1.5M`)
- [x] `TestRenderCompressionsBadge` — hidden at 0, escalates at 5/10
- [x] `TestAllocateStatusSegments_DropsInPriorityOrder` — cost drops before context bar
- [x] `TestAllocateStatusSegments_NeverTruncatesMidSegment` — atomic segment drop
- [x] `TestStaticCatalogsHaveInputLimits` — guardrail: no shipped model has `InputTokenLimit=0`
- [x] `TestStaticInputLimit` (bigmodel, moonshot) — fallback functions return correct limits
- [x] `TestOpenAILimits` — model-family limit lookup
- [x] `TestStaticLimits` (volcengine) — Doubao suffix matching
- [x] Full `go test ./internal/app/... ./internal/llm/...` green
- [x] `gofmt` / `go vet` clean
- [x] Rebased onto latest `main` (1 commit, DCO signed)

### DIFF
--- a/internal/app/conv/message.go
+++ b/internal/app/conv/message.go
@@ -37,6 +37,7 @@ type OperationModeParams struct {
 	ModelName        string
 	StatusMessage    string
 	ConversationCost llm.Money
+	Compressions     int // NEW — session compact count, drives the badge
 	Width            int
 	ThinkingEffort   string
 	ShowThinking     bool
@@ -63,7 +64,13 @@ func RenderModeStatus(params OperationModeParams) string {
 
 	left := strings.Join(leftParts, "  ")
 
-	right := renderModelWithTokens(params.ModelName, params.StatusMessage, params.InputTokens, params.InputLimit, params.ConversationCost)
+	right := renderModelWithTokens(
+		params.ModelName, params.StatusMessage,
+		params.InputTokens, params.InputLimit,
+		params.ConversationCost,
+		params.Compressions,
+		params.Width,
+	)
 	if right == "" || params.Width <= 0 {
 		return left
 	}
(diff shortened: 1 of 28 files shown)


wangke19:
/hold
need to squash commits.

wangke19:
PR is ready for review.

wangke19:
Tested with GLM,
Screenshot from 2026-06-15 19-36-22

Tested with SenseNova,
Screenshot from 2026-06-15 18-22-12

yanmxa:
Thanks so much for this — really impressive work!

internal/app/conv/status_bar.go
```diff
@@ -0,0 +1,209 @@
...
+		idx    int
+		render string
+	}
+	var survivors []kept
+	budget := availableWidth
+	for _, idx := range order {
+		s := segments[idx]
+		need := s.width
+		if len(survivors) > 0 {
+			need += len(segmentSep)
```
yanmxa, line 187: **Allocator under-counts separator width → can overflow.** Each kept segment is budgeted `len(segmentSep)` = 1 byte, but the caller in `message.go` rejoins survivors with `" · "` (3 visible cols). So the fit check is short ~2 cols per separator (up to ~10 across all segments) and the rendered cluster can exceed the budget.

Pair with the new `visualSepWidth` const (see my comment on the `segmentSep` declaration):

```suggestion
			need += visualSepWidth
```

internal/app/conv/status_bar.go
```diff
@@ -0,0 +1,209 @@
...
+	}
+	return compressionBadgeStyle(n).Render(fmt.Sprintf("🗜️ %d", n))
+}
+
+// segmentSep is the internal join separator. Uses the ASCII Unit
+// Separator (\x1f) so it can never collide with rendered segment text
+// (lipgloss emits ANSI escapes with \x1b; user strings don't contain
+// control characters). Callers split on it to swap in their own visual
+// separator.
+const segmentSep = "\x1f"
```
yanmxa, line 143: Add a column-width constant for the separator the caller actually renders, so the allocator budgets against the visible width instead of the sentinel byte length:

```suggestion
const segmentSep = "\x1f"

// visualSepWidth is the column width of the separator the caller swaps
// in (" · "). The allocator budgets against this — not len(segmentSep) —
// so the fit check matches the rendered width.
const visualSepWidth = 3
```

yanmxa (commented):
Adding inline suggestions for the points from my earlier comment. The allocator width fix (separator width + right-cluster budget) is the substantive one — the three pieces go together; the rest are minor cleanups. Thanks again for the great work! 🙏

internal/app/conv/message.go
```diff
@@ -63,7 +64,13 @@ func RenderModeStatus(params OperationModeParams) string {
 
 	left := strings.Join(leftParts, "  ")
 
-	right := renderModelWithTokens(params.ModelName, params.StatusMessage, params.InputTokens, params.InputLimit, params.ConversationCost)
+	right := renderModelWithTokens(
+		params.ModelName, params.StatusMessage,
+		params.InputTokens, params.InputLimit,
+		params.ConversationCost,
+		params.Compressions,
+		params.Width,
```
yanmxa, line 72: **Allocator gets the full line width, but the right cluster only owns `width − left − gap`.** When `left` is non-empty (accept-edits/bypass mode, thinking indicator, or queue badge), the cluster can be allocated near-full-width and overflow into the left side / wrap — exactly the narrow-terminal case the allocator targets.

Pass the right cluster's actual share instead (`-3` reserves the 2-col min gap + the `-1` in the gap formula below; `left` is already in scope):

```suggestion
		max(0, params.Width-lipgloss.Width(left)-3),
```

internal/app/conv/status_bar.go
```diff
@@ -0,0 +1,209 @@
...
+	"charm.land/lipgloss/v2"
+
+	"github.com/genai-io/san/internal/app/kit"
+)
+
+// Threshold percentages for the 4 PRD §7.2 color tiers.
+const (
+	pctGood     = 50.0
+	pctWarn     = 80.0
+	pctCritical = 95.0
```
yanmxa, line 20: Minor: this duplicates `autoCompactThreshold = 95` in `message.go`. The critical tier *means* "about to auto-compact", so they must stay equal — derive it to keep them in lockstep:

```suggestion
	pctCritical = autoCompactThreshold // critical tier == auto-compact trigger
```

internal/llm/openai/catalog.go
```diff
@@ -37,9 +37,38 @@ func openAIThinkingEfforts(model string) []string {
...
+
+// openAILimits returns known context/output windows for OpenAI model IDs.
+// OpenAI's /v1/models endpoint doesn't include limits, so we rely on
+// published specs. Returns 0,0 for unrecognized IDs.
+func openAILimits(modelID string) (input, output int) {
+	m := strings.ToLower(strings.TrimSpace(modelID))
+	switch {
+	case strings.HasPrefix(m, "gpt-6"),
+		strings.HasPrefix(m, "gpt-5.5"),
+		strings.HasPrefix(m, "gpt-5.4"):
```
yanmxa, line 58: Minor: `gpt-5.5` / `gpt-5.4` are already matched by the `gpt-5` prefix case just below (same return), so this case only needs `gpt-6`:

```suggestion
	case strings.HasPrefix(m, "gpt-6"):
```

internal/app/conv/status_bar.go
```diff
@@ -0,0 +1,209 @@
...
+	contextBarWidth = 10
+)
+
+// contextTier classifies a context-window fill percentage into one of
+// the 4 PRD §7.2 tiers. Off-by-one preserved: 80 itself falls into warn,
+// only strictly-greater-than-80 is bad. ≥95 is critical.
+type contextTier int
+
+const (
+	tierNone     contextTier = iota // pct unknown — denominator missing
```
yanmxa, line 32: Minor (no action needed): `classifyContextTier` never returns `tierNone` — the unknown-limit case is handled separately in `RenderContextBar`/`RenderContextLabel`, so this constant and its `style()` branch are currently unreachable. Drop it, or wire the unknown path through it if that was the intent.

PR #219, opened 2026-06-15
feat(plugin): manifest-style marketplaces + add-dialog paste/UI

yanmxa:
Adds support for Claude Code's manifest-style plugin marketplaces (where `marketplace.json` declares plugins whose content may live in their own repos) and fixes the Add Marketplace dialog. Previously such a marketplace showed `0 available` because plugin discovery only scanned the repo for vendored subdirectories and never read `marketplace.json`.

- **Add Marketplace dialog**: route bracketed paste (`tea.PasteMsg`) to the active overlay instead of letting it leak into the hidden prompt textarea, so the source field is pasteable; redesign the dialog (filled input box, examples-first layout).
- **Manifest discovery**: parse a plugin's `source` as either a relative-path string or a `github`/`url`/`git-subdir`/`npm` object; list and count "available" from the manifest, falling back to a directory scan for vendored marketplaces.
- **Install**: fetch external plugin sources by cloning their own repo (honoring `ref`/`sha`); resolve relative-path sources inside the marketplace repo.
- **Robustness**: self-heal a stale marketplace `installLocation` to the canonical clone path so older configs still resolve.
- **Docs**: add a marketplace & install flow diagram (mermaid) and a `source`-format table to `docs/packages/2-feature/plugin.md`, plus a co-located Chinese translation `plugin.zh.md`.

npm sources are parsed but not yet installable (returns a clear error).

### DIFF
--- a/docs/packages/2-feature/plugin.md
+++ b/docs/packages/2-feature/plugin.md
@@ -5,6 +5,8 @@ layer: feature
 
 # plugin
 
+> 中文版本：[`plugin.zh.md`](plugin.zh.md)
+
 Plugin loader, installer, marketplace, and aggregator. A plugin is a
 single package (directory) that may contribute skills, subagent
 definitions, slash commands, MCP servers, hooks, and env vars — this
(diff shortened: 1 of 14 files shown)


yanmxa:
Ready for review. ✅ build / vet / tests pass, and the install flow was verified end-to-end in the app (real `git clone` of an external plugin source, installed + enabled). A /simplify pass was applied on top (shared path guard, install-location heal moved to load, no-op cleanup removed).

PR #229, opened 2026-06-16
refactor(conv): derive pctCritical from AutoCompactThreshold

wangke19:
## Summary

Addresses minor cleanup feedback from PR #216 review.

## Changes

- **`internal/app/conv/message.go`**: Exported `AutoCompactThreshold` constant (renamed from `autoCompactThreshold` to `AutoCompactThreshold`)
- **`internal/app/conv/status_bar.go`**: Derived `pctCritical` from `AutoCompactThreshold` instead of hardcoding `95.0`

## Rationale

Eliminates duplicate constant between two files in the same package. The critical tier percentage must match the auto-compact trigger, so deriving it ensures they stay in lockstep.

## Testing

- All existing tests pass
- No behavioral changes

### DIFF
--- a/internal/app/conv/message.go
+++ b/internal/app/conv/message.go
@@ -28,6 +28,11 @@ const (
 	// agentContentIndent is the extra indent for agent prompt/response content
 	// beyond toolResultExpandedStyle's PaddingLeft(4). Total indent = 4 + 4 = 8 chars.
 	agentContentIndent = "    "
+
+	// autoCompactThreshold is the context usage percentage that triggers
+	// auto-compaction. pctCritical in status_bar.go derives from this; do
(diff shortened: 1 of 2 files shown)


yanmxa:
Thanks for the follow-up @wangke19 — nice cleanup, deriving `pctCritical` from `AutoCompactThreshold` is exactly the right call to keep the two in lockstep. 🙏

One thing before this can merge: the **DCO check is failing** because the commit isn't signed off. Could you add a `Signed-off-by` trailer matching the commit author? The quickest way:

```bash
git commit --amend -s --no-edit
git push --force-with-lease
```

That'll re-run the DCO check and turn it green. Thanks again!

wangke19:
all are green, ready for merge.

PR #231, opened 2026-06-16
fix(app): make --persona take effect in -p print mode

ldpliu:
## Summary
- When `--persona` and `-p` are combined, `runPrint` now loads the persona's identity/behavior/rules to build the system prompt via `system.Build()`
- Uses the persona's model if configured in `settings.json`
- Applies persona's `disabledTools` filtering

Previously `runPrint` ignored the persona entirely and used the hardcoded `DefaultSystemPrompt`.

## Test plan
- [x] `go build ./...` passes
- [x] `go test ./internal/app/...` passes
- [x] `go test ./internal/core/system/...` passes

### DIFF
--- a/internal/app/run.go
+++ b/internal/app/run.go
@@ -11,6 +11,7 @@ import (
 	"github.com/genai-io/san/internal/app/kit"
 	"github.com/genai-io/san/internal/app/trigger"
 	"github.com/genai-io/san/internal/core"
+	"github.com/genai-io/san/internal/core/system"
 	"github.com/genai-io/san/internal/hook"
 	"github.com/genai-io/san/internal/llm"
 	"github.com/genai-io/san/internal/persona"
@@ -27,7 +28,7 @@ func Run(opts setting.RunOptions) error {
 	}
 
 	if opts.Print != "" {
-		return runPrint(opts.Print)
+		return runPrint(opts.Print, opts.Persona)
 	}
 
 	if userQuit, err := kit.ResolveTheme(setting.LoadTheme(), setting.SaveTheme); userQuit || err != nil {
@@ -136,7 +137,7 @@ func formatAsyncHookContinuationContext(result hook.AsyncHookResult, reason stri
 	)
 }
 
-func runPrint(userMessage string) error {
+func runPrint(userMessage, personaName string) error {
 	ctx := context.Background()
 
 	llmProvider, modelID, err := resolveProvider(ctx)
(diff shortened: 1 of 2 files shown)


yanmxa (commented):
Right direction — print mode previously ignored `--persona` and used the one-line `DefaultSystemPrompt`. Two things worth tightening:

**1. Reuse the canonical tool filter, and honor global disabled tools too.**

The manual filter loop duplicates `(&tool.Set{Disabled: ...}).Tools()` (the filter used everywhere else), and it applies only the persona's `DisabledTools` — a tool the user disabled in their global `settings.json` is silently re-enabled under `-p --persona`.

The repo already has `setting.ApplyPersonaOverlay`, which merges the persona's settings over the base (model + disabledTools, correct precedence) in one place:

```go
sysPrompt := setting.DefaultSystemPrompt
var disabledTools map[string]bool
if personaName != "" {
	cwd, _ := os.Getwd()
	persona.Initialize(cwd)
	if p, ok := persona.Default().Get(personaName); ok && !p.IsBuiltin() {
		sysPrompt = system.Build(core.ScopeMain, system.WithPersona(system.Persona{
			Identity: p.Identity,
			Behavior: p.Behavior,
			Rules:    p.Rules,
		})).Prompt()

		base, _ := setting.LoadForCwd(cwd)
		var overlay *setting.Data
		if p.Settings != nil {
			overlay = &p.Settings.Data
		}
		merged := setting.ApplyPersonaOverlay(base, overlay)
		if merged.Model != "" {
			modelID = merged.Model
		}
		disabledTools = merged.DisabledTools
	}
}

schemas := (&tool.Set{Disabled: disabledTools}).Tools()
```

`Disabled: nil` returns the full set, so the no-persona path is unchanged.

**2. Model override can mismatch the connected provider (risk).**

`resolveProvider` connects a provider first, then the persona model replaces only the model id, not the provider. If the persona pins a model from a different provider, the request goes out with a model id the connected provider doesn't recognize and fails. At minimum guard it (only apply when it belongs to the connected provider, else warn/error); ideally resolve the provider from the persona's model.

**Minor:** `persona.Initialize` is already called by `validatePersona` earlier in `Run`, which also guarantees the persona exists and is non-builtin — so the second `Initialize` and the `ok && !p.IsBuiltin()` guard are redundant here (the guard's silent-fallback branch is unreachable).

ldpliu (commented):
Thanks for the thorough review!

**1. Tool filtering** — Switched to `setting.ApplyPersonaOverlay(base, overlay)` so global `disabledTools` from the user's `settings.json` are now honored alongside the persona's, and the filter uses `(&tool.Set{Disabled: disabledTools}).Tools()`.

**2. Model override guard** — Added a `ListModels` check before applying the persona's model: if the connected provider doesn't recognize it, the override is skipped to avoid a provider mismatch at request time.

**3. Redundant code** — Removed the second `persona.Initialize(cwd)` call and the `!p.IsBuiltin()` guard, since `validatePersona` already covers both.

PR #232, opened 2026-06-16
docs(design): add shared work queue ADR and update persona references

ldpliu:
## Summary
- Add ADR-0003: file-based JSONL shared work queue with `san queue` subcommand (EN + CN)
- Replace all `san agent run --persona` references with `san --persona <name> -p` in ADR-0002
- Update ADR-0002 implementation plan to reflect completed PR #231 (print mode persona fix)

## Test plan
- [x] Docs only — no code changes
- [x] Both EN and CN versions aligned

### DIFF
--- a/docs/cn/design/decisions/0002-autonomous-dev-management.md
+++ b/docs/cn/design/decisions/0002-autonomous-dev-management.md
@@ -2,7 +2,7 @@
 
 ## 状态
 
-提案中 — 2026-06-11.
+提案中 — 2026-06-11. 更新于 2026-06-16.
 
 ## 背景
 
@@ -26,14 +26,13 @@
 ```
 genai-io（Org，已存在）
 ├── san              ← San 源码仓库（已存在，被管理的目标项目）
-└── san-team         ← 新建的仓库（San 开发团队 Persona）
+└── san-team         ← 新建的仓库（纯内容，无 Go 代码）
     ├── leader/
     │   ├── system/
     │   │   ├── identity.md
     │   │   ├── behavior.md
     │   │   └── rules.md
     │   ├── skills/
-    │   │   └── ...
     │   └── settings.json
     ├── dev/
     │   ├── system/
@@ -41,7 +40,6 @@ genai-io（Org，已存在）
     │   │   ├── behavior.md
     │   │   └── rules.md
     │   ├── skills/
-    │   │   └── ...
     │   └── settings.json
     ├── qe/
     │   ├── system/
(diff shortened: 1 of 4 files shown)


yanmxa:
@ldpliu This one looks good content-wise, but the branch has fallen behind `main`. Could you rebase onto the latest `main` (or click "Update branch")? Once it's fresh and CI is green we can merge. Thanks! 🙏

ldpliu:
@yanmxa  done

docs/cn/design/decisions/0002-autonomous-dev-management.md
```diff
@@ -73,127 +71,215 @@ genai-io（Org，已存在）
...
+
+### 安装：链接 Persona
+
+启动前，将 san-team 中的 persona 链接到 San 的 persona 目录：
+
+```bash
+ln -s /path/to/san-team/leader  ~/.san/personas/leader
+ln -s /path/to/san-team/dev     ~/.san/personas/dev
+ln -s /path/to/san-team/qe      ~/.san/personas/qe
+ln -s /path/to/san-team/release ~/.san/personas/release
```
yanmxa, line 115: I'd like to keep the file not linked, how about using the settings.json to maintain the team/member

ldpliu: Which settings.json are you referring to? The per-persona settings.json inside each persona directory (e.g., san-team/leader/settings.json), or San's own ~/.san/settings.json?

If the idea is to maintain a team/member mapping inside settings.json, it would be quite difficult to configure in practice. Teams is currently a virtual concept — it's not a runtime entity in San. The san-team repo is just a directory structure that holds persona configurations. San only recognizes --persona <name> at startup; there's no --team flag and no team abstraction to resolve.

So it's hard to use a settings.json-based team/member config to launch different persona roles.

Right now persona loading only looks in two places (see internal/persona/persona.go):

1. ~/.san/personas/<name>/ — user level
2. .san/personas/<name>/ — project level (higher priority)

You either place the persona directory directly under one of these paths, or symlink it there. No other loading paths exist. The symlink approach isn't the most elegant, but it's simple and direct — no extra team abstraction layer needed.

Given these constraints, do you see a better alternative?

yanmxa:
Thanks for putting this together. The multi-agent coordination idea behind the shared work queue (ADR-0003) is genuinely interesting, but it's a big design direction and I'd rather not lock it into the repo as an ADR before we've talked through the approach. For now we want to keep things simple and not commit to a specific coordination mechanism.

I'm happy to open an issue to properly discuss the multi-agent direction — the trade-offs of a file-based JSONL queue vs. alternatives, coordination semantics, etc. — and we can revisit an ADR once there's a shared direction.

Going to close this for now. The ADR-0002 persona-reference cleanup is fine on its own, so feel free to re-send that part as a small standalone PR if you'd like it merged. Thanks again!

PR #239, opened 2026-06-17
feat(llm): add Z.ai GLM Coding Plan provider endpoint [Symphony]

wangke19:
## Summary
Add Z.ai GLM Coding Plan provider endpoint

Register a new `bigmodel:coding` auth method that uses the dedicated Coding API endpoint (`https://open.bigmodel.cn/api/coding/paas/v4`) instead of the general API endpoint.

**Files changed:**
```
 internal/llm/bigmodel/coding.go    | 39 ++++++++++++++++++++++++++++++++++++++
 internal/llm/types.go              |  1 +
 internal/setting/client_options.go |  6 ++++++
 3 files changed, 46 insertions(+)
```

**Modified files:**
```
internal/llm/bigmodel/coding.go
internal/llm/types.go
internal/setting/client_options.go
```

## Commits
```
3278980 feat(llm): add Z.ai GLM Coding Plan provider endpoint
```

## CI Results (`go build ./cmd/san/`)
```
(passed)
```

---

Created by symphony agent (issue: local-0001)

### DIFF
--- a/README.md
+++ b/README.md
@@ -157,7 +157,7 @@ Config lives in `~/.san/` (user) and `<project>/.san/` (project, overrides user)
 | **Moonshot** (Kimi) | `MOONSHOT_API_KEY` |
 | **Alibaba** (Qwen) | `DASHSCOPE_API_KEY` |
 | **MiniMax** | `MINIMAX_API_KEY` |
-| **Z.ai** (GLM) | `BIGMODEL_API_KEY` |
+| **Z.ai** (GLM / GLM Coding Plan) | `BIGMODEL_API_KEY` |
 | **SenseNova** | `SENSENOVA_API_KEY` |
 | **Mimo** | `MIMO_API_KEY` |
 | **Volcengine** (Ark) | `VOLCENGINE_API_KEY` |
--- a/docs/guides/getting-started.md
+++ b/docs/guides/getting-started.md
@@ -42,7 +42,7 @@ for Anthropic). Supported providers and the env var each one reads:
 | Alibaba | `DASHSCOPE_API_KEY` |
 | MiniMax | `MINIMAX_API_KEY` |
 | MiMo | `MIMO_API_KEY` |
-| Z.ai (GLM) | `BIGMODEL_API_KEY` |
+| Z.ai (GLM / GLM Coding Plan) | `BIGMODEL_API_KEY` |
 | DeepSeek | `DEEPSEEK_API_KEY` |
 | Ollama (local) | `OLLAMA_BASE_URL` (default `http://localhost:11434/v1`) |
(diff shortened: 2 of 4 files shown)


wangke19:
Supported coding plan of Zhipu,
image
Listed models:
image

wangke19:
Selected glm-5.2 works fine.
image

wangke19:
Ready for merge.

yanmxa:
Thanks for this! 🙏 The Coding Plan differs from the existing `api_key` method only in base URL and key — same auth, same `NewClient`, same behavior. And `BIGMODEL_BASE_URL` is already overridable, so it works today with no code:

```
BIGMODEL_API_KEY=<coding plan key>
BIGMODEL_BASE_URL=https://open.bigmodel.cn/api/coding/paas/v4
```

The only things this code adds are running both endpoints at once (with separate keys) and a ready-made UI entry. If you need either, lets keep it; otherwise I'd lean toward just documenting the `BIGMODEL_BASE_URL` override, which avoids overloading `AuthMethod` for an endpoint variant. Either way works — just want to decide deliberately.

wangke19:
@yanmxa Fair point — agreed the env var override works today with no code change. Keeping the dedicated entry though, mainly for discoverability: most users won't know that `BIGMODEL_BASE_URL` is overridable or which value to set, and a UI entry makes the Coding Plan a one-click pick instead of reading docs and editing env vars. On the `AuthMethod` overloading — agreed it's a small cost; happy to refactor to a separate `Endpoint` field on `Meta` if you'd prefer, but I think the UX win justifies it.

wangke19:
Follow-up: took another pass and simplified to reuse `BIGMODEL_API_KEY` for both Direct API and Coding Plan — the only difference is the endpoint, picked via the UI. No extra env var to manage. Kept `BIGMODEL_CODING_BASE_URL` as an optional override for advanced cases.

wangke19:
Here is the behavior from opencode, lgtm.
image

yanmxa:
Thanks for the thorough investigation here — really appreciate you weighing the trade-off rather than just defending the patch. The discoverability argument is convincing: a one-click Coding Plan entry beats expecting users to know `BIGMODEL_BASE_URL` exists and which value to set.

And the simplification is exactly right — reusing `BIGMODEL_API_KEY` for both Direct API and Coding Plan (endpoint picked via the UI) keeps it to a single key to manage, with `BIGMODEL_CODING_BASE_URL` as an optional advanced override. The `AuthCoding` method to model the endpoint variant is a fine, low-cost way to do it; no need to refactor to a separate `Endpoint` field. Cross-checking against opencode's behavior was a nice touch too.

Let's go with this approach. 👍

Issue #88, opened 2026-06-03
Improve project CI — contributor-friendly task list


yanmxa:
## Goal

Harden and modernize our CI. This is a **tracking issue** split into small, independent tasks — each checkbox is one focused PR, good for first-time contributors.

**To claim one:** comment below, open a PR, and link it here. Use Conventional Commits + `git commit -s` (DCO).

## Tasks

**PR hygiene**
- [x] **Lint PR titles** for Conventional Commits (`amannn/action-semantic-pull-request`) — _good first issue_
- [x] **Enforce DCO sign-off** so unsigned commits are caught — _good first issue_

**Speed & robustness**
- [ ] **Cache Go deps** via `setup-go` `cache: true` to cut CI time — _good first issue_
- [ ] **Add `go test -race`** to catch data races — _good first issue_
- [ ] **Adopt `golangci-lint`** to replace the bare `go vet` step — _intermediate_

**Quality**
- [ ] **Report coverage** (`-coverprofile` → Codecov), and add tests where missing — e.g. `internal/app/kit/suggest` — _intermediate_

**Security**
- [ ] **Dependabot** for the `github-actions` ecosystem — _good first issue_
- [ ] **Pin actions to SHAs** (with version comments) — _good first issue_
- [ ] **CodeQL** scanning for Go — _intermediate_

**Housekeeping**
- [ ] **`actions/stale`** to nudge/close abandoned PRs & issues — _good first issue_

## Current CI
- `ci.yml` — format check, build, unit + integration tests, `go vet`. No dependency caching.
- `pr-freshness.yml` — on push to `main`, labels/comments open PRs that are *superseded* or *conflicting*.

## Maintainer-only (not contributor tasks)
- Enable branch protection **"Require branches up to date before merging"** (or a merge queue) — the primary "stay current" mechanism; `pr-freshness` only complements it.
- Set **Actions → Workflow permissions → "Read and write"** so `pr-freshness` can comment/label.

Got another CI idea? Comment and we'll add it.

yanmxa:
Status check against current main — closing this, the tracked work is essentially done:

- [x] PR title lint + DCO — handled by genai-io/san-ci
- [x] Go dep caching — `setup-go` `cache: true`
- [x] `-race` — `make cover` runs unit + integration tests with the race detector
- [x] `golangci-lint` — `make lint` (plus layer check), `.golangci.yml`
- [x] Dependabot — `github-actions` + grouped `gomod`
[...]

PR #241, opened 2026-06-18
ci: pin remaining GitHub Actions

onyx679:
## Summary

- Pin the remaining unpinned GitHub Actions references to commit SHAs.
- Keep version comments next to each SHA for readability and Dependabot context.
- Covers the pages, PR freshness, and release workflows.

Part of genai-io/san#88 ("Pin actions to SHAs").

## Validation

- `git diff --check` passed.
- Verified no remaining unpinned `uses: ...@(vN|main|master)` entries in `.github/workflows`.
- Resolved SHAs from the upstream action tags with `git ls-remote`.

### DIFF
--- a/.github/workflows/pages.yml
+++ b/.github/workflows/pages.yml
@@ -26,10 +26,10 @@ jobs:
       url: ${{ steps.deployment.outputs.page_url }}
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v6
-      - uses: actions/configure-pages@v6
-      - uses: actions/upload-pages-artifact@v5
+      - uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10 # v6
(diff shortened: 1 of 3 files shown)


yanmxa:
Nice hardening pass — verified the SHAs and 6 of 7 are spot on. One is pinned to the wrong object:

```yaml
- uses: actions/github-script@373c709c69115d41ff229c7e5df9f8788daa9553 # v9
```

`373c709c…` is the **annotated tag object** for `v9`, not the commit it points to. SHA pins should reference the immutable commit, so this should be:

```yaml
- uses: actions/github-script@3a2844b7e9c422d3c10d287c895573f7108da1b3 # v9
```

Worth catching here because the only workflow using `github-script` (`pr-freshness.yml`) is `workflow_dispatch`-only, so CI never resolves this line. The other six (`checkout`, `setup-go`, `deploy-pages`, `action-gh-release`, `configure-pages`, `upload-pages-artifact`) all resolve correctly. 👍

yanmxa:
Close this cause the version has been updated

PR #242, opened 2026-06-18
docs(design): add default read-only persona ADR

ldpliu:
## Summary

- New ADR-0004 proposing a built-in readonly persona for San
- Overrides all three prose parts (identity, behavior, rules) with minimal prompts — ~80% token savings vs San defaults
- Two-layer defense: `permissions.deny` (enforcement) + `rules.md` (advisory)
- Two core benefits: environment protection and token savings
- Git hooks evaluated and deferred — permission engine already covers the tool surface
- Chinese translation included under `docs/cn/`

## Test plan

- [ ] Verify the ADR renders correctly in GitHub markdown preview
- [ ] Check that all internal links resolve (references to persona.md, permission-model.md, other ADRs)
- [ ] Confirm the Chinese translation is consistent with existing cn/ ADRs (0002, 0003)


yanmxa:
Thanks for this @ldpliu — a read-only persona is genuinely useful; environment protection and token savings are both real wins.

That said, I'd rather not land it as a built-in baked into core. A persona is already a self-contained, switchable folder — `.san/personas/<name>/` with `system/`, `skills/`, and `settings.json`, swapped via `/identity`. A read-only persona needs nothing new from the engine: `system/{identity,behavior,rules}.md` plus a `settings.json` with `permissions.deny` is the whole thing, and as your ADR notes the permission engine already enforces the tool surface. So there's no core change to justify, and we don't need to consolidate all this design doc into the main repo.

Concretely, I'd suggest:

1. Build it as an **actual persona folder** rather than an ADR, and
2. Publish it in its **own repo**. Then anyone who wants it just downloads the folder and enables it via `/identity` — which is exactly the distribution model personas are designed for.

If we want to show the pattern in-repo, a single small example is enough — e.g. define an `explore` / read-only persona (or agent) here to demonstrate the shape, without committing core to maintaining the full thing.

One practical note: this PR also re-touches `0002`/`0003` (EN + CN), which overlaps with your #232 and will conflict. Worth scoping each PR to its own files.

WDYT?

ldpliu:
Thanks @yanmxa — updated:

1. **Persona folder, not built-in**: revised the ADR to ship as a standard persona directory (`system/{identity,behavior,rules}.md` + `settings.json` with `permissions.deny`), no `embed.FS` or engine changes.
2. **Own repo**: published at https://github.com/genai-io/readonly-persona — installable via `git clone` into `~/.san/personas/readonly/`.
3. **Scoped the PR**: removed 0002/0003 files that belong to #232 — this PR now only contains the read-only persona ADR (0004).

PTAL

yanmxa:
One naming suggestion: consider naming this persona `explorer` rather than `readonly`.

1. **Persona convention** — San's personas are role nouns (`ml-researcher`, `social-creator`). `explorer` fits that; `readonly` names a constraint, not a role. The identity reads more naturally as "You are an explorer" than "You are a read-only assistant."
2. **Existing vocabulary** — San already has an `explore` permission mode (read-only checker) and an `Explore` subagent. This persona's `permissions.deny` is essentially explore-mode for the top-level session, so `explorer` keeps one concept family across mode / subagent / persona while the noun form marks it as the persona — whereas `readonly` adds a separate term for something San already calls explore.

(Non-blocking; the ADR itself is solid. The cn/ translation can render it 探索者.)

yanmxa:
Thanks for the thorough write-up — it's a well-reasoned proposal.

That said, we're going to pass on adding this for now. The main reason is that we want to keep the repo simple and avoid shipping built-in entities like this. Our focus right now is on opening up the underlying capability, not bundling specific personas into the repo.

If you want a read-only setup, you can already get there today: generate a persona yourself and adjust its permission config (`permissions.deny`) on top — which is exactly the enforcement layer this ADR describes. If you find yourself reusing personas like this a lot, a good pattern is to keep a personal workspace with your own repo of commonly-used personas.

Closing this out, but thanks again for taking the time to think it through.

Issue #261, opened 2026-07-01
add the new version check for the san binary


hchenxa:
### Duplicate check

- [x] I searched existing issues and discussions and didn't find a duplicate

### Problem

this is used to add a commands which can easy find the available version for binary upgrade

### Proposed solution

add a commands in cli usage to do the version update/upgrade to the latest version

hchenxa:
The PR is merged. we can close this feature request now.

PR #263, opened 2026-07-01
feat: add binary update/upgrade

hchenxa:
## What & why

https://github.com/genai-io/san/issues/261

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Refactor / cleanup
- [ ] Docs
- [ ] Other:

## Testing

```bash
```

## Checklist

- [ ] Tests pass locally

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -28,3 +28,25 @@ jobs:
         run: make cover
       - name: Vet
         run: make lint
+
+  test-windows:
+    runs-on: windows-latest
+    steps:
+      - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
+      - uses: actions/setup-go@4a3601121dd01d1626a1e23e37211e3254c1c06c # v6.4.0
+        with:
+          go-version: '1.25.8'
+          cache: true
+      - name: Build
+        run: go build ./cmd/san/
+      - name: Test
+        run: go test -count=1 ./...
+      - name: Smoke test — zip package
+        shell: bash
+        run: |
+          GOOS=windows GOARCH=amd64 go build -o san.exe ./cmd/san/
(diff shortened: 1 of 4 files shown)


yanmxa:
Thanks @hchenxa — nice addition, and the test coverage is thorough! 🙏

One thing to fix before merge:

**Cross-device rename will fail on many Linux setups.** The binary is extracted into `os.MkdirTemp("", ...)` (i.e. `$TMPDIR` / `/tmp`), then installed with `os.Rename(binPath, exe)`. When `/tmp` is a tmpfs (the systemd default on Fedora/Arch and others) and the binary lives on a different filesystem such as our default `~/.local/bin`, `os.Rename` returns `EXDEV` ("invalid cross-device link") and the update aborts. Note `install.sh` gets away with the same move only because shell `mv` falls back to copy — `os.Rename` does not.

Suggested fix: create the temp dir next to the target binary, e.g. `os.MkdirTemp(filepath.Dir(exe), ".san-update-*")`, so the final rename stays on one filesystem (with a copy fallback on `EXDEV` if you want to be extra safe).

Minor: releases ship Windows as `.zip`, but the asset name is hard-coded to `.tar.gz`, so `san update` won't find a Windows asset — worth either handling `.zip` or explicitly noting Windows isn't supported yet.

hchenxa:
@yanmxa this is a good suggestion.  and I also add the windows ci testing, please take a look at again

yanmxa (approved):
LGTM — the EXDEV cross-device fix (temp dir next to the target binary + copy fallback) and the Windows `.zip` handling with the new Windows CI job both look good, and the test coverage is thorough.

I pushed one small follow-up commit: `extractTarGz` was missing the path-traversal guard that `extractZip` already has, so a crafted tarball entry could write outside the temp dir. Added the same check plus a test. Thanks for the solid work @hchenxa!

PR #266, opened 2026-07-02
feat(desktop): add opt-in full-screen conversation surface

yanmxa:
Add an opt-in full-screen "desktop" surface — an alt-screen reader — alongside the default inline scrollback view. Ctrl-G toggles it.

## Why a second surface

A terminal program has three ways to put content on screen, and the choice decides who owns the history:

| Mode | Who owns the screen | History lives in | Scrolling |
| --- | --- | --- | --- |
| **Plain streaming** — write lines to stdout | nobody | terminal | terminal |
| **Inline** — manage the bottom N lines, push finished output above | app owns the bottom strip | terminal | terminal |
| **Alt-screen** — own every cell | app | app | app |

**San is inline today.** Bubble Tea manages the bottom strip; finished messages go above it via `tea.Println` (`insertAbove`) and become ordinary terminal scrollback. That buys native scroll, native search, native selection/copy, output that survives the session, and repaints that cost the same no matter how long the conversation is.

The flip side is that **a committed line can never be touched again**: a resize can't re-wrap it, a theme switch can't re-color it, nothing in-app can scroll or fold it, and the live region is capped at the terminal height — a long streaming reply shows only its tail until it commits.

The desktop is the alt-screen mode, offered as a second surface rather than a replacement:

- the whole conversation scrolls in-app (pgup/pgdown/home/end), independent of the terminal
- nothing is truncated — a reply taller than the screen is scrollable while it streams
- committed output stays re-renderable, so width and theme changes apply to all of it
- it's a stable full-screen frame, which is what a multi-pane layout would need

Neither mode dominates, which is why both stay. Full write-up in `docs/concepts/rendering.md`.

## Keeping it out of the inline path

- `internal/app/desktop` is a plain view component: a scrollable transcript viewport above the footer. It holds no reference to the root model and imports no bubbletea. The app hands it a `Pane`, it scrolls the result — a second window is a second Pane.
- `internal/app/desktop_surface.go` is the only seam: Surface enum, toggle, repaint heartbeat, content snapshot, key routing. Existing files gain single-line hooks (40 lines across 7 files); no inline renderer changes behavior.
- Both surfaces call the same render functions, so they read identically.
- The desktop claims only `ctrl+g` and the paging keys, routed below the overlays that already own pgup/pgdown/home/end. Everything else — typing, Enter, Ctrl-C, Ctrl-O, history — falls through unchanged.

## The one thing they can't share

`insertAbove` scrolls and inserts rows into whichever buffer is current, so a `tea.Println` issued while the alt-screen is up would land inside the desktop's frame and be lost from the history below.

So: **while the desktop owns the screen the commit pipeline runs untouched, but its terminal writes wait in the queue.** Commit offsets and `CommittedCount` advance exactly as they would inline — nothing diverges — and the queued chunks replay in order when the inline view returns.

Two consequences worth naming:

- The desktop draws the *streaming* message whole, ignoring the commit offsets that let the inline view skip its already-flushed prefix — that prefix sits in the scrollback the desktop is covering, so honouring them would open a hole mid-reply.
- Quitting from the desktop leaves the held writes unflushed, so that stretch of conversation isn't in the terminal's scrollback afterwards. It is in the session transcript, and exiting to inline first replays everything.

An overlay (slash-command picker, approval/question modal) takes the frame back for as long as it is up, so the panel that owns the keyboard is always the one drawn on screen.

### DIFF
--- a/docs/concepts/rendering.md
+++ b/docs/concepts/rendering.md
@@ -13,10 +13,94 @@ becomes state; this doc covers how state becomes characters on screen.
 > the alt-screen region. The "rendering pipeline" below is entirely
 > string composition; the terminal does the actual drawing.
 
+## Rendering modes: how a TUI can own a terminal
+
+A terminal program has three ways to put content on screen, and the
+choice decides who owns the history — the terminal, or the app:
+
+| Mode | Who owns the screen | History lives in | Scrolling |
+| --- | --- | --- | --- |
+| **Plain streaming** — write lines to stdout, no managed region | nobody | terminal | terminal |
+| **Inline** — manage the bottom N lines, push finished output above | app owns the bottom strip, terminal owns everything above | terminal | terminal |
+| **Alt-screen** — switch to the alternate buffer and own every cell | app | app | app |
+
+**San's native mode is inline**, and it is what the rest of this document
+describes. Bubble Tea manages only the bottom strip (the live tail plus
+the input area); every finished message is pushed above it with
+`tea.Println`, which is `insertAbove` — it scrolls the terminal and
+writes the lines into the terminal's own scrollback.
+
+That choice buys a lot, and all of it comes from the terminal rather
+than from code San has to write:
+
+- **The transcript is real terminal output.** Native scroll, native
+  search (⌘F), native selection and copy, and it survives the session —
+  quit San and the conversation is still in the buffer, next to the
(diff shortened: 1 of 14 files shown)


hchenxa:
@yanmxa the auto-rebase failed for this PR, you may need to fix the conflict first.

PR #290, opened 2026-07-09
feat(autopilot): Mission composer — evolving mission editor, paste, and polish

yanmxa:
Improvements to the /autopilot Mission composer:

- **Evolving mission editor.** The first message sets the mission verbatim; each later message refines the running draft via the copilot (folding the new instruction into the existing mission and replacing it) instead of piling raw turns on. With no model wired, later messages fold in verbatim so briefing still works offline. Renames the injected function reply/respond → refine throughout.
- **Paste support.** Implement `HandlePaste` so bracketed paste reaches the Mission and System Prompt editors (was dropped) — inserts at the cursor, preserves multi-line.
- **Polish.** A thin focus-accent rail down the composer's left edge so it reads as the active input field.

### DIFF
--- a/internal/app/autopilot.go
+++ b/internal/app/autopilot.go
@@ -17,7 +17,6 @@ import (
 	"go.uber.org/zap"
 
 	"github.com/genai-io/san/internal/app/conv"
-	"github.com/genai-io/san/internal/app/input"
 	"github.com/genai-io/san/internal/app/kit"
 	"github.com/genai-io/san/internal/core"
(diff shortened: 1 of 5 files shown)


yanmxa:
Superseded by the direct-editor Mission redesign shipped in v1.21.1. That version makes the Mission dialog a plain text editor — the box holds the whole mission — with `ctrl+r` refine, native paste, and the focus-accent rail, which together cover this PR's evolving-editor, paste, and polish goals. The branch also now conflicts with main. Closing as obsolete.

Issue #302, opened 2026-07-12
agent throws error some times.


hchenxa:
### Duplicate check

- [x] I searched existing issues and didn't find a duplicate
- [x] This is a functional bug, not a security vulnerability (those must go through [SECURITY.md](https://github.com/genai-io/san/blob/main/SECURITY.md))

### What happened?

Agent error: stream stalled (no data within idle timeout)

### Steps to reproduce

Agent error: stream stalled (no data within idle timeout)

### Environment

- San version:
- OS / arch:
- Go version (if building from source):
- Provider / model:

### Logs

```shell

```

yanmxa:
This should be fixed by #303, which is now merged. It'll be in the next release.

Closing as resolved — please feel free to reopen if you still hit the stream stall on the latest build. Thanks for the report!

PR #303, opened 2026-07-12
fix: agent time out issue

hchenxa:
## What & why

https://github.com/genai-io/san/issues/302

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor / cleanup
- [ ] Docs
- [ ] Other:

## Testing

```bash
```

## Checklist

- [ ] Tests pass locally

### DIFF
--- a/internal/agent/build.go
+++ b/internal/agent/build.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"strings"
+	"time"
 
 	"github.com/genai-io/san/internal/core"
 	"github.com/genai-io/san/internal/core/system"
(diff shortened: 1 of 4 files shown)


yanmxa (commented):
Two concerns:

1. **Default 60s → 180s** weakens an intentionally tight signal. `FirstChunkTimeout = 5m` already handles reasoning models thinking before the first token; once streaming, both DeepSeek and Claude extended thinking emit reasoning as chunks, so 60s of silence is usually a real stall. #302 has no repro. Prefer keeping the default and letting operators tune per-provider via the new fields.

2. **New `StreamFirstChunkTimeout` / `StreamIdleTimeout` on `BuildParams` have no callers** — no config/flag sets them. Wire them up in this PR or drop until wired.

Nit: `internal/core/agent.go:119` still says `default (60s)` — stale after this change.

hchenxa:
@yanmxa thanks for your comments, it make sense, i'd like to revert the default value changes and will introduce a new setting for streaming timeout

hchenxa:
@yanmxa please take a look at again

Issue #305, opened 2026-07-13
add session naming by commands


hchenxa:
### Duplicate check

- [x] I searched existing issues and discussions and didn't find a duplicate

### Problem

we current did not support to name/rename the session, so when I login the chat and try to select the session to resume, it's not a good user experience for now to select the session. 

### Proposed solution

we'd better to provide a commands to name/rename the session so that I can know which session I want to resume

hchenxa:
The PR is merged. we can close this ticket now.

PR #307, opened 2026-07-13
feat: add session naming features

hchenxa:
## What & why

https://github.com/genai-io/san/issues/305

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Refactor / cleanup
- [ ] Docs
- [ ] Other:

## Testing

```bash
```

## Checklist

- [ ] Tests pass locally

### DIFF
--- a/internal/app/env.go
+++ b/internal/app/env.go
@@ -66,6 +66,11 @@ type env struct {
 	// resumed configs take effect.
 	AutoPilot setting.AutoPilotSettings
 
+	// SessionName holds the custom session name set via /name. When non-empty
+	// it overrides the auto-generated title on save. Restored from persisted
+	// session metadata on resume.
+	SessionName string
+
 	// ── Cache (session-scoped) ──────────────────────────────────
 	FileCache                 *filecache.Cache
 	CachedUserInstructions    string
--- a/internal/app/input/name_command_test.go
+++ b/internal/app/input/name_command_test.go
@@ -1,0 +1,74 @@
+package input
+
+import (
+	"context"
+	"errors"
+	"strings"
+	"testing"
+)
+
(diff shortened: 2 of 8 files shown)


yanmxa (commented):
Thanks a lot for this, @hchenxa — session naming is a genuinely useful addition, and the wiring is clean end-to-end (`/name` → `renameSession` → `env.SessionName` → snapshot title, restored on resume). Nice touch bundling the `renderSession` width fix too: switching from `len()` to `lipgloss.Width()` correctly stops the `·` in the metadata (and any wide/CJK custom title) from throwing off the column alignment.

A couple of small, non-blocking notes:

- **Custom names skip the 60-char cap.** Auto-titles go through `GenerateTitle → truncateTitle` (max 60 runes), but a custom `/name` is stored verbatim. Display is safe (`renderSession` truncates), but the stored title can be arbitrarily long. Worth running the custom name through `truncateTitle` (or a length guard) for consistency.

- **`restoreSessionData` can't tell a custom name from an auto-title.** On resume it copies `Metadata.Title` into `SessionName` unconditionally. That's harmless today because `GenerateTitle` is deterministic on the first message, so re-freezing an auto-title reproduces the same value. Just flagging that if title generation ever becomes conversation-aware, resumed sessions would keep a stale title — a dedicated 'custom title' flag would future-proof it. Fine to leave for now.

Behavior, tests, and conventions all look good. 👍

hchenxa:
@yanmxa , let's try to let ci bot to auto merge the PR next time. we'd better to not to force merge it. the current logic is when the PR have `approved` + `lgtm` + `dco:yes` labels and all CI test passed, the ci bot can able to merge it automatic.

PR #323, opened 2026-07-18
fix(subagent): inherit parent reasoning effort so Codex subagents don't 400

yanmxa:
## Problem

Subagents launched against a ChatGPT subscription (Codex) provider failed with:

```
agent execution failed: LLM completion failed: infer:
POST "https://chatgpt.com/backend-api/codex/responses": 400 Bad Request
```

It's not auth (that would be 401/403) — a subagent inherits the parent's
provider instance and its OAuth token. The cause is request shape.

## Root cause

A subagent builds its client with `llm.NewClient(provider, modelID, 0)`
(`subagent/executor.go`) and nothing ever calls `SetThinkingEffort` on it, so
`ThinkingEffort == ""`. The OpenAI request builder only sends a `reasoning`
field when the effort is non-empty (`openai/client.go`). Codex models are
reasoning-only, and the `/codex/responses` endpoint rejects a reasoning-less
request with **400**. The main agent works because it sets
`EffectiveThinkingEffort()` on its client (`app/agent.go`).

## Fix

Propagate the parent's effective thinking effort to **same-provider** subagents:

- `resolveModel` now also returns `onParentProvider` (true for inherit/alias,
  false for a routed vendor).
- `buildAgent` applies the parent effort only when the run stays on the parent's
  provider — a routed subagent keeps the empty default, since the parent's
  effort may be invalid for a different vendor's model.
- `SetParentThinkingEffort` is wired from the TUI (`app/agent.go`).

Model-resolution priority is unchanged: `req.Model` → `config.Model` → inherit
the parent session's model.

## Tests

`resolveModel` tests now assert the `onParentProvider` flag (inherit/alias →
true, routed vendor → false). `go build ./...`, `go vet`, and the
`internal/subagent` / `internal/llm` / `internal/app` suites pass.

### DIFF
--- a/.san/personas/software-engineer/skills/simplify/SKILL.md
+++ b/.san/personas/software-engineer/skills/simplify/SKILL.md
@@ -32,9 +32,13 @@ in this conversation.
 
 ## Phase 2: Review across three lenses (in parallel)
 
-Use the Agent tool to launch three reviewers concurrently in one message
-(foreground). Give each the full diff. Each returns a short list of concrete,
-located findings.
+Use the Agent tool to launch three reviewers concurrently in one message. Each
+call must use `subagent_type: "code-reviewer"`, `mode: "explore"`, and foreground
+execution (omit `run_in_background` or set it to `false`). Omit `model` so
+reviewers inherit the parent model. Subagents always share the current working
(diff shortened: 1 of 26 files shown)


yanmxa:
Close this in favor of https://github.com/genai-io/san/pull/325

PR #337, opened 2026-07-19
chore: bump version to 1.21.9

yanmxa:
Bump code version to 1.21.9 and add the release changelog.


hchenxa:
@yanmxa , I think we can update the CI bot to auto merge the bump version PR, but before that, we may need to make a clear guide for the bump version PR. and may be we can asking ci bot to do the version bump as well so that we did not need owners to create this PR

PR #356, opened 2026-07-20
refactor(todo): give the two status axes distinct types

yanmxa:
A tracker item carries **two independent statuses**, and both were bare strings with colliding constant names.

| Axis | Where | Vocabulary |
|---|---|---|
| `Item.Status` | the item's own lifecycle | `pending`, `in_progress`, `completed`, `deleted` |
| `metaStatusDetail` | how the worker's background task ended | `task.TaskStatus`: `running`, `completed`, `failed`, `killed`, `stopped` |

Different questions, different vocabularies, one overlapping literal — and `CompleteWorker` uses both within eight lines:

```go
statusDetail = string(task.StatusCompleted)   // the task's outcome
...
WithStatus(StatusCompleted)                   // the item's lifecycle
```

That works only because both spell `"completed"`.

## What the missing type cost

`todo`'s constants were untyped, so an untyped constant converts to whichever side it meets. This compiled, and was **silently always false** — `task` has no `pending`:

```go
info.Status == StatusPending      // task.TaskStatus vs untyped "pending"
```

Three consecutive fixes have landed in this area (#342, #346, #348), and #215 renamed around the tangle without typing it. After this change the same line is a compile error:

```
invalid operation: info.Status == StatusPending (mismatched types task.TaskStatus and Status)
```

## What the type flushed out

Adding it turned two latent hazards into build failures:

- **`ParseStatus`** is now the single enumeration of the valid set. `tracker_update` was duplicating it in a switch that listed three of the four statuses.
- **Metadata is `map[string]any`, read back with a `.(string)` assertion**, so a typed value stored there reads back as `""` with no error anywhere. This actually broke `"interrupted"` partway through the change and `orphan_test.go` caught it. Writes now go through `setBackgroundStatusDetail`, which owns the conversion; `TestBackgroundStatusDetailRoundTripsThroughMetadata` pins the property.

## Boundaries

`transcript.TrackerItemView` keeps its plain `string` field — it is a wire contract — and the conversion is explicit at that boundary in both directions. Tool params keep arriving as JSON strings and are parsed, not cast.

**JSON output is unchanged**: a named string type marshals identically, so persisted tracker items round-trip as before.

## Verification

`go build ./...`, `go vet ./...`, `gofmt -l`, full `go test ./...` and `go test -race` over `todo`, `app`, `tool` and `session` are clean.

### DIFF
--- a/internal/app/conv/tracker_view.go
+++ b/internal/app/conv/tracker_view.go
@@ -142,7 +142,7 @@ func renderItem(t *todo.Item, phase itemPhase, width, idWidth int, blockers func
 	switch phase {
 	case itemAborted:
 		abortedStyle := lipgloss.NewStyle().Foreground(kit.CurrentTheme.Error)
-		detail := mutedStyle.Render("[" + todo.BackgroundStatusDetail(t) + "]")
+		detail := mutedStyle.Render("[" + string(todo.BackgroundStatusDetail(t)) + "]")
 		return renderItemLine(indent, abortedStyle.Render("!"), idTag, subject, detail)
 
 	case itemFinished:
--- a/internal/app/conv/tracker_view_test.go
+++ b/internal/app/conv/tracker_view_test.go
@@ -7,6 +7,7 @@ import (
 	"strings"
 	"testing"
 
+	"github.com/genai-io/san/internal/task"
 	"github.com/genai-io/san/internal/todo"
 )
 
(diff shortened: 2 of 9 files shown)


yanmxa:
Closing on review: this fixes no reachable bug.

Every status comparison on main is single-axis and correct — no site compares an Item.Status (lifecycle) against a task.TaskStatus (terminal outcome). The 'lost interrupted marker' defect the PR cites does not exist on main: StatusDetailInterrupted is a plain string there, stored and read back as a string. That defect only comes into being *inside this diff*, which retypes the constant and then adds setBackgroundStatusDetail to convert it back.

Preventive type-safety presented as a bug fix. If the typing is wanted it should be re-filed honestly as hygiene, not merged as a fix.

PR #368, opened 2026-07-20
fix(worktree): make the git calls cancellable

yanmxa:
All four git invocations used `exec.Command` with no context:

```go
exec.Command("git", "worktree", "add", "--detach", worktreePath)   // :46
exec.Command("git", "worktree", "remove", "--force", worktreePath) // :66
exec.Command("git", "status", "--porcelain")                       // :77
exec.Command("git", "log", "--oneline", "HEAD", "--not", ...)      // :88
```

and both tool entry points **explicitly discarded** theirs:

```go
func (t *EnterWorktreeTool) Execute(_ context.Context, ...)
func (t *ExitWorktreeTool) Execute(_ context.Context, ...)
```

So pressing Esc did nothing — the agent goroutine sat in `cmd.Output()` until git returned on its own, and the turn could not be interrupted. There was no timeout anywhere either.

## Not a theoretical wait

- `git worktree add` runs the repo's `post-checkout` hook.
- `git status --porcelain` blocks on an lfs smudge filter, a network filter driver, or another process holding the index lock.

## The fix

Thread `ctx` through `Create`, `Remove`, `HasUncommittedChanges` and `HasUnmergedCommits`, and reconnect the `ctx` the two tools were dropping. `session/env.go:11` already uses `exec.CommandContext` for its git call; this brings the package in line.

`Create`'s cleanup closure gets its **own** bounded context rather than the caller's — a cancelled turn is exactly when the worktree needs removing, so inheriting an already-done ctx would leave it on disk.

## Tests

`TestGitCallsHonourACancelledContext` passes an already-cancelled context (the state after Esc) and asserts `Create`/`Remove` error rather than running, and that the two boolean predicates — which fail closed rather than returning an error — return promptly instead of blocking on git. `TestCreateAndRemoveWithALiveContext` covers the normal path end to end.

`make lint`, `gofmt -l` and the full `go test ./...` are clean.

### DIFF
--- a/internal/tool/mode/enterworktree.go
+++ b/internal/tool/mode/enterworktree.go
@@ -57,8 +57,8 @@ func (t *EnterWorktreeTool) ExecuteWithResponse(_ context.Context, params map[st
 }
 
 // Execute is the non-interactive fallback (auto-approve for subagents).
-func (t *EnterWorktreeTool) Execute(_ context.Context, params map[string]any, cwd string) toolresult.ToolResult {
-	result, cleanup, err := worktree.Create(cwd, tool.GetString(params, "name"))
+func (t *EnterWorktreeTool) Execute(ctx context.Context, params map[string]any, cwd string) toolresult.ToolResult {
+	result, cleanup, err := worktree.Create(ctx, cwd, tool.GetString(params, "name"))
 	if err != nil {
 		return toolresult.NewErrorResult(t.Name(), err.Error())
 	}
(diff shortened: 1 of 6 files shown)


yanmxa:
Closing on review: the motivating scenario is unreachable.

EnterWorktree is an InteractiveTool, and for the main agent the adapter takes the interactive branch and asserts the request as *QuestionRequest — which *EnterWorktreeRequest is not — so it errors at adapter.go:152 before any git runs. worktree.Create/Remove only execute for background subagents (askFn == nil). So 'user presses Esc while git hangs' cannot happen on the main agent, and Esc cancels the main turn, not a detached subagent.

The change is correct and harmless, but its stated severity is unreachable, and the real worktree defect (main-agent EnterWorktree is broken entirely) is elsewhere and untouched. Not worth merging as filed; the main-agent breakage deserves its own issue.

PR #371, opened 2026-07-20
fix(cron): stop one san window deleting another's scheduled jobs

yanmxa:
The durable-job file is **per-project, not per-process**: `<project>/.san/scheduled_tasks.json`. Two san windows open on the same repo write it.

`LoadDurable` runs once, at startup (`app/init.go:54`), and `saveDurableLocked` rewrote the file from the in-memory view. So a job created in one window was simply absent from the other's view, and the other's next save erased it from disk.

```
window A:  /loop 10m check the tests   -> file becomes [X]
window B:  any durable change          -> file becomes [B's jobs], X gone
```

Any durable change triggers that save — including a job **firing or expiring** (`store.go:166`, `:184`). Neither user has to touch the vanished job for it to go.

## The fix

A save now carries over the durable jobs on disk that this process has never held. Two things it deliberately is **not**:

**Not a merge into memory.** Their schedule state belongs to the instance that owns them, and adopting them into `s.jobs` would put them through `LoadDurable`'s boot-time `NextFire` recalculation. That recalculation is right for a restart — it stops a long-lived schedule replaying missed intervals — but applied mid-session it pushes a recurring job's next fire out on every save, and the job never runs.

**Not a merge with disk.** The obvious "re-read and union" resurrects a job this window deliberately deleted, because disk still has it. `knownDurable` records every durable id this process has held, which is what separates *"I removed this"* from *"another instance created it"*.

## Not addressed

**Double firing.** Two windows that were both running when a job was created each hold it and each tick it, so its prompt is injected twice. Fixing that means sharing fire state across processes, which needs the schedule-ownership question above answered first — a bigger design call than this data-loss fix, and separable from it.

## Tests

`TestSavePreservesAnotherInstancesJob` runs two schedulers on one file and asserts neither erases the other — reverting the fix fails it with *"window A's job was erased by window B's save"*. `TestDeleteIsNotUndoneByThePreservation` pins that a delete is not resurrected, `TestDeleteAfterRestartStillRemoves` covers a job adopted at boot and then removed, and `TestSessionJobsStayOutOfTheSharedFile` checks session-only jobs never reach it.

`make lint`, `gofmt -l` and the full `go test ./...` are clean.

### DIFF
--- a/internal/cron/cross_instance_test.go
+++ b/internal/cron/cross_instance_test.go
@@ -1,0 +1,137 @@
+package cron
+
+import (
+	"encoding/json"
+	"os"
+	"path/filepath"
+	"testing"
+)
+
+// instanceOn models a san process attached to a project's storage file.
+func instanceOn(t *testing.T, path string) *Scheduler {
+	t.Helper()
+	s := NewScheduler()
+	s.SetStoragePath(path)
+	if err := s.LoadDurable(); err != nil {
+		t.Fatalf("LoadDurable: %v", err)
+	}
+	return s
+}
+
+func durableIDsOnDisk(t *testing.T, path string) map[string]bool {
+	t.Helper()
+	data, err := os.ReadFile(path)
+	if err != nil {
+		if os.IsNotExist(err) {
+			return map[string]bool{}
+		}
+		t.Fatalf("read storage: %v", err)
+	}
+	var jobs []*Job
+	if err := json.Unmarshal(data, &jobs); err != nil {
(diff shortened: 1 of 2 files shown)


yanmxa:
LGTM for preventing data loss. One question: since durable jobs are project-scoped, do we intend them to be shared across sessions? This preserves jobs written by another instance but intentionally does not synchronize them into memory, while duplicate firing remains possible. If project-wide shared scheduling is the intended model, we may need a follow-up design for ownership and atomic claiming/leases.

PR #379, opened 2026-07-22
feat(bash): show a live output counter on running commands

yanmxa:
Stacked on #378 — the diff below includes that PR's timer commit until it merges; this PR's own change is the single commit for the counter. Kept as a draft; will rebase onto main and mark ready once #378 lands.

The elapsed timer (#378) says a command is alive but not whether it is progressing: a healthy build and a hung one both just tick upward. Count a foreground command's output as it streams and show it next to the timer (` · 12s · 1.2k lines`), so climbing lines mean progress and a frozen counter under a rising clock means stuck.

- `cmd.Run` is untouched: stdout/stderr are teed through a byte/line counter that forwards output verbatim, so timeout, process-group teardown, and truncation are all preserved.
- No output yet (a progress bar, one long token) falls back to a byte size.
- Snapshots are throttled and ride the existing agent→UI channel as a replaceable value; zero cost when nothing is listening.

### DIFF
--- a/internal/agent/build.go
+++ b/internal/agent/build.go
@@ -57,6 +57,7 @@ type BuildParams struct {
 	HookEngine       hook.Handler
 	AskUser          tool.AskUserFunc
 	ToolActivity     func(toolCallID string, msg string)
+	ToolProgress     func(toolCallID string, lines int, bytes int64)
 	// BashPromptResponder answers prompts a command raises *while it runs*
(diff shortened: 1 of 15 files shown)


yanmxa:
Folded into #378 — consolidating the timer and the output counter into a single PR at the author's request. The counter commit now sits on top of the timer commit on that branch.

PR #393, opened 2026-07-23
fix(agent): split stop control from spawn

yanmxa:
## Summary

- split background-agent cancellation out of `Agent` into an optional `AgentStop` tool
- disable `AgentStop` by default so spawning agents never sees a stop-only enum field
- let background Bash work report a Unix process-group ID and tested `kill` commands for graceful or forced shutdown

## Root cause

`Agent` combined spawning with cancellation through an optional `signal: "stop"` field. Models could emit that singleton enum while trying to spawn a worker, sending a placeholder task ID to the stop path and producing `task not found` instead of starting the worker.

## Behavior

`AgentStop` accepts only a running agent task ID and rejects Bash tasks. It is parent-only and disabled by default; enable it with `/tool` when agent-managed cancellation is needed. Bash remains responsible for stopping its own background commands. On Unix the reported `kill -TERM -- -<pgid>` command stops the process group and records the task as `stopped`.

## Validation

`GOCACHE=/private/tmp/san-go-build-cache go test ./...`

### DIFF
--- a/docs/guides/writing-a-subagent.md
+++ b/docs/guides/writing-a-subagent.md
@@ -128,7 +128,7 @@ A background subagent registers with the broker while running:
   on its next step. Delivery is best-effort.
 - **Report to main**: `SendMessage(to="main", message)` sends an interim note.
   The final answer returns automatically on completion.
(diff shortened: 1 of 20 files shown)


zhujian7:
/lgtm
@yanmxa please sign off the commit

PR #402, opened 2026-07-23
refactor(agent): use an implicit default subagent

yanmxa:
Replace predefined agent types with one implicit default subagent while preserving custom user, project, and plugin definitions.

- make `subagent_type` optional and reserve it for custom definitions
- enforce explore mode as read-only while allowing classified shell inspection
- keep the simplify review flow on one batch of three default reviewers
- support headless default subagent runs and inherited model validation


yanmxa:
Superseded by the focused stacked PRs:

- #407 — Agent core selection/runtime contract
- #408 — custom Agent loading, management, integrations, and docs
- #409 — MCP lifecycle/concurrency and per-Agent MCP integration

PR #404, opened 2026-07-24
feat(llm): add user-defined OpenAI/Claude-compatible custom provider

skeeey:
Adds a "Custom" provider for any OpenAI-compatible endpoint, configured in-place from the /model Providers tab via a two-field form (base URL + API key) rather than the single API-key input built-in providers use.

Supported cases:
1. Using a token-billing platform API (e.g. an OpenAI-compatible gateway).
2. Using a self-deployed LLM (e.g. vLLM, LocalAI, or any OpenAI-compatible server).

- internal/llm/custom: new provider with fixed ID "custom"; baseURL read lazily from the store on connect, API key from the secret store under SAN_CUSTOM_API_KEY. ListModels pulls context_length variants into InputTokenLimit.
- internal/llm/store: persist CustomProviderConfig (ID + baseURL) and expose CustomProvider / SetCustomProvider.
- internal/app/input: inline form (on_provider_custom.go) with Tab/arrow field switching, paste support, validation, and a connect flow that refreshes the Providers tab; routed through nav, view, auth, and data lifecycle hooks.
- internal/app/run: DefaultModel now keys on the bare provider name (provider names carry name:auth) and requires an explicit selection when no default applies (e.g. the custom provider).

## What & why

Closes #

## Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Refactor / cleanup
- [ ] Docs
- [ ] Other:

## Testing

```bash
```

## Checklist

- [ ] Tests pass locally

### DIFF
--- a/internal/app/input/on_provider_custom.go
+++ b/internal/app/input/on_provider_custom.go
@@ -96,7 +96,7 @@ func (s *ProviderSelector) focusCustomFormField(i int) {
 // validateCustomForm normalizes and validates the two fields. An empty API key
 // means "keep the stored one" and only fails when nothing is stored.
 func (s *ProviderSelector) validateCustomForm() (baseURL, apiKey string, err error) {
-	baseURL = strings.TrimSuffix(strings.TrimSpace(s.customFormInputs[customFormFieldBaseURL].Value()), "/")
+	baseURL = strings.TrimRight(strings.TrimSpace(s.customFormInputs[customFormFieldBaseURL].Value()), "/")
 	if baseURL == "" {
 		return "", "", fmt.Errorf("base URL is required")
 	}
--- a/internal/app/input/on_provider_custom.go
+++ b/internal/app/input/on_provider_custom.go
@@ -1,0 +1,155 @@
+// Provider selector: the custom (third-party, OpenAI-compatible) provider form.
+// Built-in providers only need an API key, so Enter opens the single inline
+// input; the custom provider also needs a baseURL, so it gets a two-field form
+// instead. The provider ID is fixed (custom.DefaultID) — there is only one
+// custom provider, so a user-chosen ID would add rename bookkeeping without
+// distinguishing anything.
+package input
+
+import (
+	"fmt"
+	"os"
+	"strings"
+
+	"charm.land/bubbles/v2/textinput"
+	tea "charm.land/bubbletea/v2"
+
+	"github.com/genai-io/san/internal/llm"
+	"github.com/genai-io/san/internal/llm/custom"
+	"github.com/genai-io/san/internal/secret"
+)
+
+// customFormField indexes customFormInputs.
+const (
+	customFormFieldBaseURL = iota
+	customFormFieldAPIKey
+	customFormFieldCount
+)
+
+// isCustomProvider reports whether name is the custom provider's fixed name.
+func (s *ProviderSelector) isCustomProvider(name llm.Name) bool {
+	return name == llm.Name(custom.DefaultID)
+}
+
+// openCustomForm initializes the inline form, prefilled from the saved config.
+// The API key field starts empty like the built-in edit flow; on submit an
+// empty key keeps the stored one.
+func (s *ProviderSelector) openCustomForm() {
+	baseURL := ""
+	if s.store != nil {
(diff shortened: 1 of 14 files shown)


internal/setting/client_options.go
```diff
@@ -32,6 +32,8 @@ func DefaultModel(providerName string, authMethod string) string {
 	case "sensenova":
 		return "sensenova-6.7-flash-lite"
 	default:
-		return "claude-sonnet-4-20250514"
+		// Unknown provider (e.g. a user-defined custom provider) — there is no
+		// sensible default; the caller must require an explicit selection.
+		return ""
```
hchenxa, line 48: **DefaultModel returns "" for ollama, mimo, volcengine**  
Three built-in providers (ollama, mimo, volcengine) don't have cases in `DefaultModel`'s switch. The `strings.Cut` fix in `run.go` correctly normalizes the name for the 10 providers that DO have cases, but the 3 missing ones previously got the fallback `"claude-sonnet-4-20250514"` (wrong model, but no error) and now get `""`, which triggers the new hard error in `resolveProvider`:

> no model selected for ollama:api_key. Run 'san' and use /model to select one

**When it bites:** A user runs `san -p` (headless/print mode) with an ollama/mimo/volcengine provider connected but no saved current model. The interactive TUI is fine — it saves the model on use.

**Fix:** Add cases for ollama, mimo, and volcengine to the switch, or better, drive the switch off the provider registry so it stays in sync automatically.

internal/app/input/on_provider_custom.go
```diff
@@ -0,0 +1,155 @@
...
+func (s *ProviderSelector) focusCustomFormField(i int) {
+	s.customFormInputs[s.customFormFocus].Blur()
+	s.customFormFocus = i
+	s.customFormInputs[i].Focus()
+}
+
+// validateCustomForm normalizes and validates the two fields. An empty API key
+// means "keep the stored one" and only fails when nothing is stored.
+func (s *ProviderSelector) validateCustomForm() (baseURL, apiKey string, err error) {
+	baseURL = strings.TrimSuffix(strings.TrimSpace(s.customFormInputs[customFormFieldBaseURL].Value()), "/")
```
hchenxa: **`strings.TrimSuffix` only removes a single trailing slash**  
If a user enters a base URL like `https://api.example.com/v1//` (two trailing slashes), `strings.TrimSuffix(s, "/")` leaves one slash intact, producing `https://api.example.com/v1/`. Downstream, the openai-go SDK may construct paths like `https://api.example.com/v1//chat/completions`, which some servers reject as non-canonical.

**Fix:** Use `strings.TrimRight(s, "/")` to remove all trailing slashes, or better, use `url.JoinPath()` for proper URL normalization.

hchenxa (commented):
## Review: Custom provider feature

The implementation is solid and well-tested. Two findings below.

---
@skeeey 
 
Two findings: one medium (DefaultModel regression for ollama/mimo/volcengine) and one low (TrimSuffix only removes a single trailing slash).

yanmxa:
Thanks @skeeey ! ! !

@zhujian7 Is this the implemetation for your proposal?

zhujian7:
@yanmxa yes, customize the base url and token.
Thanks @skeeey !

internal/setting/client_options.go
skeeey: Fixed in 01fcb44. Added explicit cases to `DefaultModel`: `llama4` for ollama and `xiaomi/mimo-v2.5-pro` for mimo (both from their catalogs). Volcengine has no static catalog — the model comes from `VOLCENGINE_MODEL` — so the case now resolves that env var, and only errors when it's unset (which genuinely requires an explicit selection). Covered by a new `client_options_test.go`.

internal/app/input/on_provider_custom.go
skeeey: Fixed in 01fcb44 — switched to `strings.TrimRight(..., "/")` so repeated trailing slashes are all removed, and extended the form validation test to cover the `v1//` case. Went with TrimRight over `url.JoinPath` since the base URL is stored as a prefix for the SDK, not joined with paths here.

PR #407, opened 2026-07-24
feat(agent): restore custom agent runtime

yanmxa:
Restore the custom Agent selection and runtime contract while keeping the implicit default Agent.

- Resolve exact enabled Agent names before permission or execution.
- Apply Agent-specific model, mode, step, prompt, skill, and tool policies.
- Preserve named Agent identity across foreground and background execution.

Stack: **1/3** → #408 → #409


yanmxa:
Superseded by the clean net-diff stack #410 → #411 → #412. The replacement removes the delete-and-restore review churn.

PR #410, opened 2026-07-24
feat(agent): simplify agent runtime selection

yanmxa:
Replace bundled Agent types with one implicit default Agent while preserving exact custom Agent selection.

- Resolve and pin named Agent configurations before permission approval.
- Apply model, permission, prompt, skill, step, and tool policies at runtime.
- Preserve Agent identity across foreground, background, task, and hook paths.

Stack: **1/3** → #411 → #412


internal/tool/agent/schema.go
```diff
@@ -66,29 +66,21 @@ var agentToolParameters = map[string]any{
 			"type":        "string",
 			"description": "A short (3-5 word) description of the task",
 		},
-		"subagent_type": map[string]any{
-			"type":        "string",
-			"description": "The type of specialized agent to use for this task",
-		},
 		"name": map[string]any{
```
hchenxa, line 67: **`name` now selects the custom agent instead of serving as a display label**  
This is a clean simplification — the old schema had both `subagent_type` (which agent to use) and `name` (a display label), which was confusing. Now `name` is the custom agent selector, and the display label is inferred from the mode (Explorer/Editor) or the agent name itself. The implicit default path (`name=""`) maps correctly to the unlabeled default.  
  
One note for backward compatibility: old tool invocations in session history that used `subagent_type` will replay with `name=""`, silently falling back to the default agent instead of whatever named agent was originally selected. Since the system prompt is also updated this shouldn't cause divergence in practice.

internal/tool/agent/schema.go
```diff
@@ -66,29 +66,21 @@ var agentToolParameters = map[string]any{
...
 			"description": "Set to true to run this agent in the background. You will be notified when it completes.",
 		},
-		"model": map[string]any{
-			"type":        "string",
-			"description": "Optional model override. If omitted or unavailable, inherits from parent conversation.",
-		},
 		"max_steps": map[string]any{
 			"type":        "number",
-			"description": "Maximum number of LLM inference steps for the agent. Built-in agents default to 100 and lower values are raised to 100.",
+			"description": "Maximum number of LLM inference steps. Defaults to 500; lower values are raised to 500.",
```
hchenxa, line 77: **`max_steps` default and minimum raised from 100 to 500**  
The description now explicitly says "lower values are raised to 500" and `types.go` documents the constant as "both the default AND minimum." For custom agent definitions that previously worked with fewer steps (e.g. a lightweight reviewer with `MaxSteps=50`), this is a behavioral change — they'll now run up to 500 steps. Worth a mention in release notes for anyone with custom agent configs.

internal/subagent/executor.go
```diff
@@ -562,6 +688,20 @@ func subagentPermissionFunc(mode PermissionMode, allowRules, denyRules ToolList)
 		if tool.IsParentOnlyTool(name) {
 			return false, fmt.Sprintf("tool %s is reserved for the main conversation", name)
 		}
+		// Explore is an authoritative read-only boundary. Explicit allow rules
```
hchenxa, line 680: **Explore mode as an authoritative read-only ceiling**  
The explore gate inserted before allow_tools is the right design — a custom agent definition can narrow its tool set within explore mode, but cannot accidentally elevate to workspace mutation through allow_tools. The suite of denial tests in `TestExploreModeCannotElevateThroughAllowTools` covers all the smuggling vectors (Write, Edit, `mkdir`, `npm install`, `git commit`, and even `git diff --output`), which gives good confidence this boundary holds.

hchenxa (commented):
## Review: feat(agent): simplify agent runtime selection

Solid refactor. The implicit default agent with `name`-based custom selection is cleaner than the old built-in catalog, and the permission model changes are well-considered and thoroughly tested.

A few observations inline below — all design-level notes rather than bugs.

PR #412, opened 2026-07-24
fix(mcp): manage per-agent server lifecycles

yanmxa:
Make MCP connections safe for concurrent per-Agent use.

- Add lease- and generation-aware connection ownership and cleanup.
- Preserve persistent intent and callbacks across reconnect, adoption, and reload.
- Connect configured MCP servers for TUI and headless Agent runs.

Stack: #410 → #411 → **3/3**

### DIFF
--- a/internal/mcp/caller_test.go
+++ b/internal/mcp/caller_test.go
@@ -1,0 +1,293 @@
+package mcp
+
+import (
+	"context"
+	"encoding/json"
+	"sync"
+	"testing"
+	"time"
+
+	"github.com/genai-io/san/internal/mcp/transport"
+)
+
+func TestLeaseReleaseDoesNotDisconnectPreexistingRetainedConnection(t *testing.T) {
+	tr := newSlowTransport(0)
+	registry := connectedRegistry(t, map[string]transport.Transport{"shared": tr})
+
+	cleanup, errs := AcquireServerConnectionLeases(context.Background(), registry, []string{"shared"})
+	if len(errs) != 0 {
+		t.Fatalf("AcquireServerConnectionLeases() errors = %v", errs)
+	}
+	cleanup()
+
+	if _, ok := registry.GetClient("shared"); !ok {
+		t.Fatal("cleanup disconnected a connection owned by another caller")
+	}
+}
+
+type connectionLifecycleTransport struct {
+	mu         sync.Mutex
+	alive      bool
+	closeCalls int
+}
+
+func (t *connectionLifecycleTransport) Start(context.Context) error {
+	t.mu.Lock()
+	t.alive = true
+	t.mu.Unlock()
+	return nil
+}
+
+func (t *connectionLifecycleTransport) Send(_ context.Context, req *transport.JSONRPCRequest) (*transport.JSONRPCResponse, error) {
+	result := json.RawMessage(`{}`)
+	return &transport.JSONRPCResponse{JSONRPC: "2.0", ID: req.ID, Result: result}, nil
+}
+
+func (t *connectionLifecycleTransport) SendNotification(context.Context, *transport.JSONRPCNotification) error {
+	return nil
+}
+
+func (t *connectionLifecycleTransport) Close() error {
+	t.mu.Lock()
+	t.alive = false
+	t.closeCalls++
+	t.mu.Unlock()
+	return nil
+}
+
+func (t *connectionLifecycleTransport) IsAlive() bool {
+	t.mu.Lock()
+	defer t.mu.Unlock()
+	return t.alive
+}
+
+func (t *connectionLifecycleTransport) SetNotificationHandler(transport.NotificationHandler) {}
+
+func (t *connectionLifecycleTransport) closeCount() int {
+	t.mu.Lock()
+	defer t.mu.Unlock()
+	return t.closeCalls
+}
+
+func TestConcurrentLeaseAcquisitionSharesOneConnectionUntilFinalRelease(t *testing.T) {
+	registry := NewRegistryForTest(map[string]ServerConfig{
+		"shared": {Name: "shared", Type: TransportSTDIO, Command: "shared"},
+	})
+	tr := &connectionLifecycleTransport{}
+	var factoryCalls int
+	registry.newClientForConfig = func(cfg ServerConfig) *Client {
+		factoryCalls++
+		client := NewClient(cfg)
+		client.TransportFactory = func() (transport.Transport, error) { return tr, nil }
+		return client
+	}
+
+	start := make(chan struct{})
+	cleanups := make(chan func(), 2)
+	errors := make(chan []error, 2)
+	var wg sync.WaitGroup
+	for range 2 {
+		wg.Add(1)
+		go func() {
+			defer wg.Done()
+			<-start
+			cleanup, errs := AcquireServerConnectionLeases(context.Background(), registry, []string{"shared"})
+			cleanups <- cleanup
+			errors <- errs
+		}()
+	}
+	close(start)
+	wg.Wait()
+
+	for range 2 {
+		if errs := <-errors; len(errs) != 0 {
+			t.Fatalf("AcquireServerConnectionLeases() errors = %v", errs)
+		}
+	}
+	if factoryCalls != 1 {
+		t.Fatalf("client factory calls = %d, want 1", factoryCalls)
+	}
+
+	firstCleanup := <-cleanups
+	secondCleanup := <-cleanups
+	firstCleanup()
+	if _, ok := registry.GetClient("shared"); !ok {
+		t.Fatal("first cleanup disconnected a connection still leased by another caller")
+	}
+	if tr.closeCount() != 0 {
+		t.Fatal("first cleanup closed the shared transport")
+	}
+
+	secondCleanup()
+	if _, ok := registry.GetClient("shared"); ok {
+		t.Fatal("last cleanup left its temporary connection in the registry")
+	}
+	deadline := time.Now().Add(time.Second)
+	for tr.closeCount() == 0 && time.Now().Before(deadline) {
+		time.Sleep(time.Millisecond)
+	}
+	if tr.closeCount() != 1 {
+		t.Fatalf("transport close count = %d, want 1", tr.closeCount())
+	}
+}
+
+func TestExplicitConnectRetainsLeaseCreatedConnectionAfterFinalRelease(t *testing.T) {
+	registry := NewRegistryForTest(map[string]ServerConfig{
+		"shared": {Name: "shared", Type: TransportSTDIO, Command: "shared"},
+	})
+	tr := &connectionLifecycleTransport{}
+	registry.newClientForConfig = func(cfg ServerConfig) *Client {
+		client := NewClient(cfg)
+		client.TransportFactory = func() (transport.Transport, error) { return tr, nil }
+		return client
+	}
+
+	cleanup, errs := AcquireServerConnectionLeases(context.Background(), registry, []string{"shared"})
+	if len(errs) != 0 {
+		t.Fatalf("AcquireServerConnectionLeases() errors = %v", errs)
+	}
+	if err := registry.Connect(context.Background(), "shared"); err != nil {
+		t.Fatalf("Connect() error = %v", err)
+	}
+	cleanup()
+
+	if _, ok := registry.GetClient("shared"); !ok {
+		t.Fatal("lease cleanup disconnected a connection promoted by an explicit Connect")
+	}
+	if tr.closeCount() != 0 {
+		t.Fatal("lease cleanup closed a connection promoted by an explicit Connect")
+	}
+}
+
+func TestLeaseSetReleaseIsIdempotent(t *testing.T) {
+	registry := NewRegistryForTest(map[string]ServerConfig{
+		"shared": {Name: "shared", Type: TransportSTDIO, Command: "shared"},
+	})
+	tr := &connectionLifecycleTransport{}
+	registry.newClientForConfig = func(cfg ServerConfig) *Client {
+		client := NewClient(cfg)
+		client.TransportFactory = func() (transport.Transport, error) { return tr, nil }
+		return client
+	}
+
+	first, errs := AcquireServerConnectionLeases(context.Background(), registry, []string{"shared"})
+	if len(errs) != 0 {
+		t.Fatalf("first AcquireServerConnectionLeases() errors = %v", errs)
+	}
+	second, errs := AcquireServerConnectionLeases(context.Background(), registry, []string{"shared"})
+	if len(errs) != 0 {
+		t.Fatalf("second AcquireServerConnectionLeases() errors = %v", errs)
+	}
+
+	first()
+	first()
+	if _, ok := registry.GetClient("shared"); !ok {
+		t.Fatal("repeated cleanup released another caller's lease")
+	}
+	second()
+	waitFor(t, "the final lease cleanup to close the transport", func() bool { return tr.closeCount() == 1 })
+}
+
+func TestDeadConnectionReplacementKeepsLeaseCountsSeparatedByConnectionEpoch(t *testing.T) {
+	registry := NewRegistryForTest(map[string]ServerConfig{
+		"shared": {Name: "shared", Type: TransportSTDIO, Command: "shared"},
+	})
+	firstTransport := &connectionLifecycleTransport{}
+	secondTransport := &connectionLifecycleTransport{}
+	transports := []transport.Transport{firstTransport, secondTransport}
+	registry.newClientForConfig = func(cfg ServerConfig) *Client {
+		tr := transports[0]
+		transports = transports[1:]
+		client := NewClient(cfg)
+		client.TransportFactory = func() (transport.Transport, error) { return tr, nil }
+		return client
+	}
+
+	first, errs := AcquireServerConnectionLeases(context.Background(), registry, []string{"shared"})
+	if len(errs) != 0 {
+		t.Fatalf("first AcquireServerConnectionLeases() errors = %v", errs)
+	}
+	firstTransport.mu.Lock()
+	firstTransport.alive = false
+	firstTransport.mu.Unlock()
+	second, errs := AcquireServerConnectionLeases(context.Background(), registry, []string{"shared"})
+	if len(errs) != 0 {
+		t.Fatalf("replacement AcquireServerConnectionLeases() errors = %v", errs)
+	}
+
+	first()
+	second()
+	if _, ok := registry.GetClient("shared"); ok {
+		t.Fatal("generation-specific lease cleanup leaked the replacement connection")
+	}
+	waitFor(t, "the replacement transport to close", func() bool { return secondTransport.closeCount() == 1 })
+}
+
+func TestExplicitRetentionIntentAppliesToLeaseTriggeredReplacement(t *testing.T) {
+	registry := NewRegistryForTest(map[string]ServerConfig{
+		"shared": {Name: "shared", Type: TransportSTDIO, Command: "shared"},
+	})
+	firstTransport := &connectionLifecycleTransport{}
+	secondTransport := &connectionLifecycleTransport{}
+	transports := []transport.Transport{firstTransport, secondTransport}
+	registry.newClientForConfig = func(cfg ServerConfig) *Client {
+		tr := transports[0]
+		transports = transports[1:]
+		client := NewClient(cfg)
+		client.TransportFactory = func() (transport.Transport, error) { return tr, nil }
+		return client
+	}
+
+	if err := registry.Connect(context.Background(), "shared"); err != nil {
+		t.Fatalf("persistent Connect() error = %v", err)
+	}
+	firstTransport.mu.Lock()
+	firstTransport.alive = false
+	firstTransport.mu.Unlock()
+	cleanup, errs := AcquireServerConnectionLeases(context.Background(), registry, []string{"shared"})
+	if len(errs) != 0 {
+		t.Fatalf("AcquireServerConnectionLeases() errors = %v", errs)
+	}
+	cleanup()
+
+	if _, ok := registry.GetClient("shared"); !ok {
+		t.Fatal("lease cleanup disconnected a replacement with persistent ownership")
+	}
+	if secondTransport.closeCount() != 0 {
+		t.Fatal("lease cleanup closed a persistently owned replacement")
+	}
+}
+
+func TestOldEpochLeaseReleaseCannotDisconnectCurrentConnection(t *testing.T) {
+	registry := NewRegistryForTest(map[string]ServerConfig{
+		"shared": {Name: "shared", Type: TransportSTDIO, Command: "shared"},
+	})
+	first := &connectionLifecycleTransport{}
+	second := &connectionLifecycleTransport{}
+	transports := []transport.Transport{first, second}
+	registry.newClientForConfig = func(cfg ServerConfig) *Client {
+		tr := transports[0]
+		transports = transports[1:]
+		client := NewClient(cfg)
+		client.TransportFactory = func() (transport.Transport, error) { return tr, nil }
+		return client
+	}
+
+	cleanup, errs := AcquireServerConnectionLeases(context.Background(), registry, []string{"shared"})
+	if len(errs) != 0 {
+		t.Fatalf("AcquireServerConnectionLeases() errors = %v", errs)
+	}
+	registry.Disconnect("shared")
+	if err := registry.Connect(context.Background(), "shared"); err != nil {
+		t.Fatalf("replacement Connect() error = %v", err)
+	}
+	cleanup()
+
+	client, ok := registry.GetClient("shared")
+	if !ok || client == nil {
+		t.Fatal("stale lease cleanup removed the replacement connection")
+	}
+	if !second.IsAlive() {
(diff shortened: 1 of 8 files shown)


hchenxa:
## PR Review: `fix(mcp): manage per-agent server lifecycles`

### Overview

This PR (+675/−120, 8 files) introduces a lease-based connection ownership model for MCP servers, enabling safe concurrent per-Agent use. The lease tracking, epoch comparisons, and lock ordering are all **correct** — no correctness bugs were found. The code is well-tested with 7 new test cases.

However, there are several design concerns worth addressing before merging.

---

### 1. `connectionLifecycleMu` held across network I/O — Medium severity

**File:** `internal/mcp/registry.go` — `Connect` / `acquireConnectionLease` → `ensureConnectedWhileLifecycleLocked`

`connectionLifecycleMu` is acquired at the start of `Connect`/`acquireConnectionLease` and not released until after `ensureConnectedWhileLifecycleLocked` returns. Inside that function, `client.Connect(ctx)` is called — which starts a transport subprocess and performs a synchronous JSON-RPC initialize handshake. Both are blocking I/O operations that can take seconds for remote MCP servers (SSE, streamable HTTP).

During this time, **every** other lifecycle operation is blocked — TUI `/mcp` disconnect, other agents lease acquisition, `Reload`, `RemoveServer`, `DisconnectAll` — because they all contend on the same `connectionLifecycleMu`. The old code ran `client.Connect(ctx)` outside any mutex.

**Recommendation:** Release `connectionLifecycleMu` before calling `client.Connect()` and reacquire it afterward. The duplicate-connection race this prevents also exists in the old code and is handled correctly by the `r.clients[name]` check-then-set pattern (last writer wins; connections are stateless with respect to the registry).

---

### 2. `RemoveServer` holds `connectionLifecycleMu` during synchronous `client.Disconnect()` — Medium severity

**File:** `internal/mcp/registry.go:275–284`

`RemoveServer` calls `client.Disconnect()` synchronously while holding both `connectionLifecycleMu` and `r.mu`. `Client.Disconnect` waits for the transport read loop to exit (up to 2s) and the subprocess to terminate (up to 5s) — see `transport/stdio.go:262–284`. This blocks **all** lifecycle operations for up to 7 seconds with a wedged server.

This contradicts the pattern established in PR #367 (commit `039d4229`), which specifically moved `Disconnect`s teardown to a background goroutine to prevent this exact UI freeze. `RemoveServer` was never updated to match, and PR #412 adds `connectionLifecycleMu` on top, making the freeze global.

**Failure scenario:** A user removes a hung MCP server via the TUI. For up to 7 seconds, no other server can connect, disconnect, or have its leases managed. The `/mcp` panel appears frozen.

**Recommendation:** Extract the client from `r.clients` under the lock, release both locks, then call `disconnectClientAsync(name, client)` — matching the pattern used by `Disconnect`:

```go
r.connectionLifecycleMu.Lock()
r.mu.Lock()
client, ok := r.clients[name]
if ok {
    delete(r.clients, name)
    r.connectionEpoch[name]++
    // delete other maps...
}
r.mu.Unlock()
r.connectionLifecycleMu.Unlock()
if ok {
    disconnectClientAsync(name, client)
    r.notifyToolsChanged()
}
```

---

### 3. Four parallel maps should be a per-server struct — Low severity (maintenance risk)

**File:** `internal/mcp/registry.go` — `Registry` struct fields

The PR adds four parallel maps keyed by server name:

```go
leaseCountsByConnectionEpoch map[string]map[uint64]int
disconnectAfterFinalLease    map[string]bool
retainWithoutLeases          map[string]bool
connectionEpoch              map[string]uint64
```

Every cleanup site (`Reload`, `Disconnect`, `DisconnectAll`, `RemoveServer`, `transferRetainedConnectionsFrom`, `releaseConnectionLease`) must remember to update all four maps plus `clients`, `connecting`, and `connectErr`. Currently **3 sites** handle all four in sequential lines; **2 sites** (`RemoveServer` and `transferRetainedConnectionsFrom`) already have omissions where one or more maps are skipped.

A single `map[string]*serverConnectionState` struct would reduce cleanup to a single `delete()` and make omissions impossible:

```go
type serverConnectionState struct {
    leaseCountsByEpoch       map[uint64]int
    disconnectAfterFinalLease bool
    retainWithoutLeases       bool
    currentEpoch              uint64
}
```

---

### 4. `Reload` operator precedence makes condition fragile — Low severity

**File:** `internal/mcp/registry.go:230`

```go
if !r.retainWithoutLeases[name] || ok && reflect.DeepEqual(config, r.configs[name]) {
```

Due to Go operator precedence (`&&` before `||`), this evaluates as `(!retain) || (ok && deepEqual)`. The logic is correct, but the lack of parentheses makes the intent non-obvious. Adding `( )` would prevent future misinterpretation.

---

### 5. Dead nil-map guards obscure construction gap — Low severity

**File:** `internal/mcp/registry.go:328` and `:386`

`ensureConnectedWhileLifecycleLocked` and `acquireConnectionLease` contain nil-guards for `connectionEpoch` and `leaseCountsByConnectionEpoch`. All three constructors (`newEmptyRegistry`, `NewRegistryForTest`, `NewRegistry`) already initialize these maps, so the guards are dead code. The real issue is that the three constructors independently duplicate initialization — a new map field added to one but not the others would cause a nil-map panic rather than a compile-time error.

---

### 6. Manual unlocks without `defer` in `releaseConnectionLease` — Very low severity

**File:** `internal/mcp/registry.go` — `releaseConnectionLease`

Two early return paths manually unlock both `r.mu` and `r.connectionLifecycleMu`. A future edit adding a new early return that misses one of the unlocks would cause a deadlock. Consider wrapping the `mu`-protected section in an anonymous function with `defer`.

---

### Summary

| # | Finding | Severity |
|---|---|---|
| 1 | `connectionLifecycleMu` held across `client.Connect()` network I/O | Medium |
| 2 | `RemoveServer` synchronous `Disconnect` under `connectionLifecycleMu` | Medium |
| 3 | Four parallel maps should be a per-server struct | Low |
| 4 | `Reload` opaque precedence | Low |
| 5 | Dead nil-map guards obscure init gap | Low |
| 6 | Manual unlocks without `defer` in `releaseConnectionLease` | Very Low |

No correctness bugs found — lease epoch tracking, lock ordering (`connectionLifecycleMu → mu`), and ownership flags are all implemented correctly with good test coverage.

hchenxa (changes requested):
## PR Review: `fix(mcp): manage per-agent server lifecycles`

### Overview

This PR (+675/−120, 8 files) introduces a lease-based connection ownership model for MCP servers, enabling safe concurrent per-Agent use. The lease tracking, epoch comparisons, and lock ordering are all **correct** — no correctness bugs were found. The code is well-tested with 7 new test cases.

However, there are several design concerns worth addressing before merging.

---

### 1. `connectionLifecycleMu` held across network I/O — Medium severity

**File:** `internal/mcp/registry.go` — `Connect` / `acquireConnectionLease` → `ensureConnectedWhileLifecycleLocked`

`connectionLifecycleMu` is acquired at the start of `Connect`/`acquireConnectionLease` and not released until after `ensureConnectedWhileLifecycleLocked` returns. Inside that function, `client.Connect(ctx)` is called — which starts a transport subprocess and performs a synchronous JSON-RPC initialize handshake. Both are blocking I/O operations that can take seconds for remote MCP servers (SSE, streamable HTTP).

During this time, **every** other lifecycle operation is blocked — TUI `/mcp` disconnect, other agents lease acquisition, `Reload`, `RemoveServer`, `DisconnectAll` — because they all contend on the same `connectionLifecycleMu`. The old code ran `client.Connect(ctx)` outside any mutex.

**Recommendation:** Release `connectionLifecycleMu` before calling `client.Connect()` and reacquire it afterward. The duplicate-connection race this prevents also exists in the old code and is handled correctly by the `r.clients[name]` check-then-set pattern (last writer wins; connections are stateless with respect to the registry).

---

### 2. `RemoveServer` holds `connectionLifecycleMu` during synchronous `client.Disconnect()` — Medium severity

**File:** `internal/mcp/registry.go:275–284`

`RemoveServer` calls `client.Disconnect()` synchronously while holding both `connectionLifecycleMu` and `r.mu`. `Client.Disconnect` waits for the transport read loop to exit (up to 2s) and the subprocess to terminate (up to 5s) — see `transport/stdio.go:262–284`. This blocks **all** lifecycle operations for up to 7 seconds with a wedged server.

This contradicts the pattern established in PR #367 (commit `039d4229`), which specifically moved `Disconnect` teardown to a background goroutine to prevent this exact UI freeze. `RemoveServer` was never updated to match, and PR #412 adds `connectionLifecycleMu` on top, making the freeze global.

**Failure scenario:** A user removes a hung MCP server via the TUI. For up to 7 seconds, no other server can connect, disconnect, or have its leases managed. The `/mcp` panel appears frozen.

**Recommendation:** Extract the client from `r.clients` under the lock, release both locks, then call `disconnectClientAsync(name, client)` — matching the pattern used by `Disconnect`:

```go
r.connectionLifecycleMu.Lock()
r.mu.Lock()
client, ok := r.clients[name]
if ok {
    delete(r.clients, name)
    r.connectionEpoch[name]++
    // delete other maps...
}
r.mu.Unlock()
r.connectionLifecycleMu.Unlock()
if ok {
    disconnectClientAsync(name, client)
    r.notifyToolsChanged()
}
```

---

### 3. Four parallel maps should be a per-server struct — Low severity (maintenance risk)

**File:** `internal/mcp/registry.go` — `Registry` struct fields

The PR adds four parallel maps keyed by server name:

```go
leaseCountsByConnectionEpoch map[string]map[uint64]int
disconnectAfterFinalLease    map[string]bool
retainWithoutLeases          map[string]bool
connectionEpoch              map[string]uint64
```

Every cleanup site (`Reload`, `Disconnect`, `DisconnectAll`, `RemoveServer`, `transferRetainedConnectionsFrom`, `releaseConnectionLease`) must remember to update all four maps plus `clients`, `connecting`, and `connectErr`. Currently **3 sites** handle all four in sequential lines; **2 sites** (`RemoveServer` and `transferRetainedConnectionsFrom`) already have omissions where one or more maps are skipped.

A single `map[string]*serverConnectionState` struct would reduce cleanup to a single `delete()` and make omissions impossible:

```go
type serverConnectionState struct {
    leaseCountsByEpoch       map[uint64]int
    disconnectAfterFinalLease bool
    retainWithoutLeases       bool
    currentEpoch              uint64
}
```

---

### 4. `Reload` operator precedence makes condition fragile — Low severity

**File:** `internal/mcp/registry.go:230`

```go
if !r.retainWithoutLeases[name] || ok && reflect.DeepEqual(config, r.configs[name]) {
```

Due to Go operator precedence (`&&` before `||`), this evaluates as `(!retain) || (ok && deepEqual)`. The logic is correct, but the lack of parentheses makes the intent non-obvious. Adding `( )` would prevent future misinterpretation.

---

### 5. Dead nil-map guards obscure construction gap — Low severity

**File:** `internal/mcp/registry.go:328` and `:386`

`ensureConnectedWhileLifecycleLocked` and `acquireConnectionLease` contain nil-guards for `connectionEpoch` and `leaseCountsByConnectionEpoch`. All three constructors (`newEmptyRegistry`, `NewRegistryForTest`, `NewRegistry`) already initialize these maps, so the guards are dead code. The real issue is that the three constructors independently duplicate initialization — a new map field added to one but not the others would cause a nil-map panic rather than a compile-time error.

---

### 6. Manual unlocks without `defer` in `releaseConnectionLease` — Very low severity

**File:** `internal/mcp/registry.go` — `releaseConnectionLease`

Two early return paths manually unlock both `r.mu` and `r.connectionLifecycleMu`. A future edit adding a new early return that misses one of the unlocks would cause a deadlock. Consider wrapping the `mu`-protected section in an anonymous function with `defer`.

---

### Summary

| # | Finding | Severity |
|---|---|---|
| 1 | `connectionLifecycleMu` held across `client.Connect()` network I/O | Medium |
| 2 | `RemoveServer` synchronous `Disconnect` under `connectionLifecycleMu` | Medium |
| 3 | Four parallel maps should be a per-server struct | Low |
| 4 | `Reload` opaque precedence | Low |
| 5 | Dead nil-map guards obscure init gap | Low |
| 6 | Manual unlocks without `defer` in `releaseConnectionLease` | Very Low |

No correctness bugs found — lease epoch tracking, lock ordering (`connectionLifecycleMu → mu`), and ownership flags are all implemented correctly with good test coverage.

internal/mcp/registry.go
```diff
@@ -32,19 +32,26 @@ type PluginServer struct {
...
-	loader     *ConfigLoader
-	cwd        string
+	mu                           sync.RWMutex
+	connectionLifecycleMu        sync.Mutex
+	clients                      map[string]*Client
+	configs                      map[string]ServerConfig
+	disabled                     map[string]bool           // servers explicitly disabled by the user
+	connecting                   map[string]bool           // servers currently being connected (async)
+	connectErr                   map[string]string         // last connection error for servers without a client
+	leaseCountsByConnectionEpoch map[string]map[uint64]int // active per-agent users grouped by connection generation
```
hchenxa, line 42: **Four parallel maps should be a per-server struct**

These four maps (`leaseCountsByConnectionEpoch`, `disconnectAfterFinalLease`, `retainWithoutLeases`, `connectionEpoch`) are all keyed by server name and always managed together. Currently **3 sites** (`Reload`, `Disconnect`, `DisconnectAll`) handle all four in sequential lines, and **2 sites** (`RemoveServer` at line 295-297, `transferRetainedConnectionsFrom` at line 102-125) already have gaps where one or more maps are omitted.

A single `map[string]*serverConnectionState` struct would reduce cleanup to a single `delete()` and make omissions impossible:

```go
type serverConnectionState struct {
    leaseCountsByEpoch       map[uint64]int
    disconnectAfterFinalLease bool
    retainWithoutLeases       bool
    currentEpoch              uint64
}
```

internal/mcp/registry.go
```diff
@@ -160,18 +201,36 @@ func (r *Registry) Reload() error {
...
+	// lease-owned clients stay with their current registry until their owner
+	// releases them; reload must not invalidate an in-flight Agent run.
 	for name, client := range r.clients {
-		if _, ok := configs[name]; !ok {
-			_ = client.Disconnect()
-			delete(r.clients, name)
-			delete(r.connecting, name)
-			delete(r.connectErr, name)
+		config, ok := configs[name]
+		if !r.retainWithoutLeases[name] || ok && reflect.DeepEqual(config, r.configs[name]) {
```
hchenxa, line 212: **Operator precedence makes condition fragile**

Due to Go operator precedence (`&&` before `||`), this evaluates as `(!retain) || (ok && deepEqual)`. The logic is correct, but adding parentheses would make the intent obvious at a glance and prevent future misinterpretation.

internal/mcp/registry.go
```diff
@@ -213,13 +272,16 @@ func (r *Registry) AddServer(name string, config ServerConfig, scope Scope) erro
...
 // RemoveServer removes a server configuration
 func (r *Registry) RemoveServer(name string) error {
+	r.connectionLifecycleMu.Lock()
+	defer r.connectionLifecycleMu.Unlock()
 	r.mu.Lock()
 	defer r.mu.Unlock()
 
 	// Disconnect if connected (best-effort; errors are ignored during cleanup)
 	if client, ok := r.clients[name]; ok {
 		_ = client.Disconnect()
```
hchenxa, line 282: **`RemoveServer` holds `connectionLifecycleMu` during synchronous `client.Disconnect()`**

`client.Disconnect()` waits for the transport read loop to exit (up to 2s) and the subprocess to terminate (up to 5s) — see `transport/stdio.go:262-284`. During this time, both `connectionLifecycleMu` and `r.mu` are held, blocking ALL lifecycle operations for up to 7 seconds with a wedged server.

This directly contradicts the pattern established by PR #367 (commit `039d4229`), which specifically moved `Disconnect`'s teardown to a background goroutine (`disconnectClientAsync`) to prevent this exact UI freeze. `RemoveServer` should follow the same pattern: extract the client from `r.clients` under the lock, release both locks, then call `disconnectClientAsync(name, client)`.

internal/mcp/registry.go
```diff
@@ -230,40 +292,136 @@ func (r *Registry) RemoveServer(name string) error {
...
 	}
+	factory := r.newClientForConfig
 	r.mu.Unlock()
 
-	// Create and connect client
 	client := NewClient(config)
+	if factory != nil {
+		client = factory(config)
+	}
 	if err := client.Connect(ctx); err != nil {
```
hchenxa, line 341: **`connectionLifecycleMu` held during network I/O**

`connectionLifecycleMu` is acquired at line 304 in `Connect` (or line 360 in `acquireConnectionLease`) and not released until after `ensureConnectedWhileLifecycleLocked` returns. But `client.Connect(ctx)` here performs network I/O — starting a transport subprocess and a synchronous JSON-RPC initialize handshake — while the caller's `connectionLifecycleMu` is still held.

This means **every** other lifecycle operation (TUI `/mcp` disconnect, other agents' lease acquisition, `Reload`, `RemoveServer`, `DisconnectAll`) is blocked for the duration of the connection handshake. The old code ran `client.Connect(ctx)` outside any mutex.

**Recommendation:** Release `connectionLifecycleMu` before calling `client.Connect()` and reacquire afterward. The duplicate-connection race this prevents is the same one the old code lived with, handled correctly by the `r.clients[name]` check-then-set pattern (last writer wins).

internal/mcp/registry.go
```diff
@@ -230,40 +292,136 @@ func (r *Registry) RemoveServer(name string) error {
...
+		return false, fmt.Errorf("failed to connect to %s: %w", serverName, err)
 	}
-
-	// Set up tools changed callback
 	client.SetOnToolsChanged(r.notifyToolsChanged)
 
 	r.mu.Lock()
-	r.clients[name] = client
+	r.clients[serverName] = client
+	if r.connectionEpoch == nil {
```
hchenxa, line 348: **Dead nil-map guard — all constructors initialize this map**

`connectionEpoch` is initialized by all three constructors (`newEmptyRegistry:148`, `NewRegistryForTest:164`, `NewRegistry:185`). This nil-check can never fire under any valid construction path, making it dead code. (Same issue at `acquireConnectionLease` line 368 for `leaseCountsByConnectionEpoch`.)

The root cause is that the three constructors independently duplicate map initialization. A shared base constructor would eliminate both the dead guards and the risk of a future constructor missing a new map.

internal/mcp/registry.go
```diff
@@ -230,40 +292,136 @@ func (r *Registry) RemoveServer(name string) error {
...
+// connection whose final lease ended. A stale cleanup from an older connection
+// generation cannot affect a replacement connection with the same name.
+func (r *Registry) releaseConnectionLease(serverName string, leasedConnectionEpoch uint64) {
+	r.connectionLifecycleMu.Lock()
+
+	var clientDetachedAfterFinalLease *Client
+
+	r.mu.Lock()
+	leaseCountsByEpoch := r.leaseCountsByConnectionEpoch[serverName]
+	if leaseCountsByEpoch == nil || leaseCountsByEpoch[leasedConnectionEpoch] == 0 {
```
hchenxa, line 398: **Manual unlocks without `defer`**

Both early return paths (line 399-401 and line 405-407) manually unlock `r.mu` and `r.connectionLifecycleMu`. A future edit adding a new early return that misses one of the unlocks would cause a deadlock. Consider wrapping the `mu`-protected section in an anonymous function with `defer r.mu.Unlock()`, keeping only `defer r.connectionLifecycleMu.Unlock()` at the outer level.

hchenxa (changes requested):
## PR Review: `fix(mcp): manage per-agent server lifecycles`

### Overview

This PR (+675/−120, 8 files) introduces a lease-based connection ownership model for MCP servers, enabling safe concurrent per-Agent use. The lease tracking, epoch comparisons, and lock ordering are all **correct** — no correctness bugs were found. The code is well-tested with 7 new test cases.

However, there are several design concerns worth addressing before merging.

### Summary

| # | Finding | Severity |
|---|---|---|
| 1 | `connectionLifecycleMu` held across `client.Connect()` network I/O | Medium |
| 2 | `RemoveServer` synchronous `Disconnect` under `connectionLifecycleMu` | Medium |
| 3 | Four parallel maps should be a per-server struct | Low |
| 4 | `Reload` opaque precedence | Low |
| 5 | Dead nil-map guards obscure init gap | Low |
| 6 | Manual unlocks without `defer` in `releaseConnectionLease` | Very Low |

internal/mcp/caller_test.go
```diff
@@ -0,0 +1,293 @@
...
+		t.Fatalf("AcquireServerConnectionLeases() errors = %v", errs)
+	}
+	cleanup()
+
+	if _, ok := registry.GetClient("shared"); !ok {
+		t.Fatal("cleanup disconnected a connection owned by another caller")
+	}
+}
+
+type connectionLifecycleTransport struct {
```
hchenxa, line 28: **Duplicate test helper — `connectionLifecycleTransport` vs `liveTransport`**

This new struct duplicates the existing `liveTransport` in `adopt_clients_test.go` (same package). Both implement the full `transport.Transport` interface with near-identical internals (`mu`, alive/closed state, locked accessors). The key difference is that `connectionLifecycleTransport` adds a `closeCalls` counter and returns a real JSON-RPC response in `Send` (needed because `Client.Connect` performs an initialize handshake).

Consider embedding `liveTransport` and overriding only `Close` (to increment closeCalls and delegate) and `Send` (to provide the JSON-RPC response):

```go
type connectionLifecycleTransport struct {
    liveTransport
    closeCalls int
}

func (t *connectionLifecycleTransport) Send(_ context.Context, req *transport.JSONRPCRequest) (*transport.JSONRPCResponse, error) {
    return &transport.JSONRPCResponse{JSONRPC: "2.0", ID: req.ID, Result: json.RawMessage(`{}`)}, nil
}

func (t *connectionLifecycleTransport) Close() error {
    t.liveTransport.Close()
    t.closeCalls++
    return nil
}
```

This eliminates ~30 lines of duplicated method stubs.

internal/subagent/executor.go
```diff
@@ -45,7 +45,7 @@ type Executor struct {
 	projectInstructions        string               // project memory (CLAUDE.md/AGENTS.md) for edit-capable subagents
 	skillsPrompt               string               // available skills section for capable subagents
 	mcpTools                   mcp.Tools            // tool schemas + execution
-	mcpServers                 mcp.Servers          // connect/disconnect for per-subagent server sets
+	mcpConnectionRegistry      *mcp.Registry        // per-Agent connection leases
```
hchenxa, line 48: **Tighter coupling: interface → concrete type**

The old `SetMCP` accepted `mcp.Servers` (interface), allowing any implementation. The new `SetMCPDependencies` takes `*mcp.Registry` (concrete type). This makes it harder to test or mock the executor independently — tests now need a real `*mcp.Registry` rather than a mock `Servers`.

Since the lease operations (`acquireConnectionLease`/`releaseConnectionLease`) are unexported methods on `*Registry` and don't exist on the `Servers` interface, the concrete type is necessary. Consider whether the `Servers` interface should gain these methods, or if the concrete dependency is acceptable for this project's testing patterns.

internal/mcp/registry.go
```diff
@@ -59,11 +66,12 @@ type mcpState struct {
 // guarding is the swap, which Initialize performs on the bubbletea goroutine
 // (reloadProjectServices) while DefaultRegistry() is read elsewhere.
 var (
-	registryMu      sync.RWMutex
-	defaultRegistry = newEmptyRegistry()
+	registryMu         sync.RWMutex
+	registryTransferMu sync.Mutex
```
hchenxa, line 70: **`registryTransferMu` is a redundant global lock**

This package-level `sync.Mutex` is only used in `transferRetainedConnectionsFrom` to prevent concurrent transfers. However, `connectionLifecycleMu` on each registry already serializes lifecycle operations on that registry, and the function acquires both outgoing and incoming `connectionLifecycleMu` locks. The only call site is `Initialize` in `service.go`, which runs on the single bubbletea goroutine.

The scenario `registryTransferMu` prevents — a concurrent transfer while another operation modifies the same registries — is already prevented by the per-registry `connectionLifecycleMu` locks. If kept, it should at least have a comment explaining what it protects that the per-registry locks do not.

internal/mcp/registry.go
```diff
@@ -74,32 +82,53 @@ var (
...
+	outgoing.connectionLifecycleMu.Lock()
+	defer outgoing.connectionLifecycleMu.Unlock()
+	incoming.connectionLifecycleMu.Lock()
+	defer incoming.connectionLifecycleMu.Unlock()
+
+	outgoing.mu.Lock()
+	connectionsToTransfer := make(map[string]*Client)
+	retainedConnectionsToDisconnect := make(map[string]*Client)
+	for name, client := range outgoing.clients {
+		cfg, stillConfigured := incoming.configs[name]
```
hchenxa, line 101: **Unprotected read of `incoming.configs` while holding only `outgoing.mu`**

This function reads `incoming.configs[name]` while only holding `outgoing.mu` (not `incoming.mu`). At the sole call site in `service.go:Initialize`, this is safe because `incoming` has not yet been published as the default registry, so no other goroutine can see it. However, if this function were ever called from another context where `incoming` is visible to concurrent goroutines, this would be a data race. A comment documenting this constraint would prevent future misuse.

hchenxa (commented):
Additional findings in other files:

PR #414, opened 2026-07-24
fix(agent): honor disabled tools and avoid unnecessary delegation

yanmxa:
Honor global disabled-tool settings in subagents and make the Agent tool less eager to delegate simple work.

- Remove disabled tools from subagent schemas and executors
- Prefer direct work for clear, bounded tasks
- Remove the duplicate project-level simplify skill


internal/subagent/executor.go
```diff
@@ -146,6 +150,21 @@ func (e *Executor) SetMCP(tools mcp.Tools, servers mcp.Servers) {
 	e.mcpServers = servers
 }
 
+// SetDisabledTools supplies the effective global disabled-tool policy inherited
+// by subsequently built subagents. The map is copied so settings reloads and
+// concurrent runs cannot mutate a tool set while it is being constructed.
```
hchenxa, line 158: **Inconsistent cloning:** `maps.Clone(disabled)` correctly prevents shared-map mutation, but `allow` and `disallow` slices are passed by reference into `tool.Set`. If a caller retains and mutates the returned slice later, the tool set changes unexpectedly.

Currently the callers (`rc.config.AllowTools.Names()` and `rc.config.DenyTools.BareNames()`) build fresh slices each call, so this is safe today. But the inconsistency with the cloned `disabled` map is a maintenance trap — consider cloning all three inputs or documenting the ownership contract.

internal/app/agent.go
```diff
@@ -404,6 +404,9 @@ func (m *model) ensureAgentSession(pendingSend string) (tea.Cmd, error) {
 		m.StopAgentSession()
 	}
 
+	// Rebuild the subagent executor alongside the main agent so it inherits the
+	// same effective disabled-tool snapshot after settings/persona drift.
+	m.ReconfigureAgentTool()
```
hchenxa, line 409: **Unconditional rebuild on every turn:** `ReconfigureAgentTool()` recreates the entire executor and adapter on every `ensureAgentSession` call (every user message), even if no settings changed. While the cost per message is negligible (no I/O), a diff-based check (only reconfigure when `disabledTools` actually changed) would make this more robust and avoid unnecessary allocations.

internal/subagent/executor_test.go
```diff
@@ -15,6 +15,7 @@ import (
 	"github.com/genai-io/san/internal/setting"
 	"github.com/genai-io/san/internal/skill"
 	"github.com/genai-io/san/internal/tool"
+	_ "github.com/genai-io/san/internal/tool/registry"
```
hchenxa: **Side-effect import:** `_ "github.com/genai-io/san/internal/tool/registry"` registers tools so `tool.Set.Tools()` returns non-empty schemas. Without a comment, a future reader or automatic import cleaner may not understand why this import exists. Consider adding:

```go
_ "github.com/genai-io/san/internal/tool/registry" // registers Agent, Bash, etc. for tool.Set.Tools()
```

hchenxa (commented):
## PR Review: `fix(agent): honor disabled tools and avoid unnecessary delegation`

### Overview

This PR (+117/−77, 7 files) ensures subagents inherit the main agent's disabled-tool policy, removes the duplicate project-level simplify skill, and updates the Agent tool schema to prefer direct work over delegation. The core mechanism (threading `Disabled` map through `tool.Set` and the executor) is implemented correctly — no correctness bugs found.

### Findings

| # | Finding | Severity |
|---|---|---|
| 1 | `newAgentToolSet` clones the disabled map but passes allow/disallow by reference | Low |
| 2 | Unconditional `ReconfigureAgentTool` on every `ensureAgentSession` | Low |
| 3 | Side-effect import in test file lacks explanatory comment | Very Low |

internal/subagent/executor.go
yanmxa: Fixed in b3dbf202: newAgentToolSet now clones allow and disallow with slices.Clone, and a regression test mutates all three caller-owned inputs.

internal/app/agent.go
yanmxa: No code change here: ensureAgentSession already returns early while the active session snapshots match, so ReconfigureAgentTool only runs on cold start or after capability/disabled-tool drift forces a rebuild. It is not called on every user message.

internal/subagent/executor_test.go
yanmxa: Fixed in b3dbf202 by documenting that the blank import registers the built-in schemas used by tool.Set.Tools().

PR #417, opened 2026-07-24
fix(agent): use unknown names as display labels

yanmxa:
## Summary

- treat unknown Agent `name` values as display-only labels backed by the default agent template
- continue using enabled named definitions and reject definitions that exist but are disabled
- clarify the Agent tool schema and cover foreground/background resolution behavior

## Testing

- `go test ./...`
- `go vet ./...`
- `git diff --check`


yanmxa:
LGTM

PR #426, opened 2026-07-26
docs(site): rebuild the homepage around 三

yanmxa:
The site still sold the previous positioning ("the open harness for terminal
agents") on a SaaS-shaped page: emoji feature cards, hover lifts, a stack of
teal glows, and no mention of the ~2.3k-token harness number the README now
leads with. This brings the site up to the current README and gives it one
coherent surface.

## Homepage

- **Hero** carries the README line — *Minimal harness. Maximum agent.* — with a
  metric strip and a tabbed installer (curl / powershell / go install).
- **The animated intro** gets its own full-width screen right under the hero,
  wider than the reading column, so it is watchable rather than decorative.
- **Why San** is now the trigram: 一 small, 二 fast, 三 open — one horizontal
  band each, ruled off by hairlines, prose plus mono spec rows. The third
  stroke opens into *plug in · write · oversee*. The six emoji feature cards and
  the level track fold into these three bands.
- **Benchmark** gains the row the README added — harness context ~2.3k vs
  ~20.9k tokens, ~9x leaner — and a one-sentence footnote.
- The filled-teal "try it in 30 seconds" box is gone; the hero installer
  already does that job, and removing it closes the page's largest dead gap.
- Numbers appear once each: prose states the idea, spec rows and the benchmark
  table carry the figures.

## Design system (shared by every page)

- Prose is serif throughout (Newsreader); IBM Plex Mono is reserved for the
  machine voice — nav, buttons, code, data, labels. Spline Sans is dropped, so
  one less font request.
- Parchment and teal now match `intro.html` 1:1, so the page and the embedded
  intro are the same material. The radial glow stack is replaced by the intro's
  own paper grain.
- Hairline rules, soft corners, no hover lifts. Emoji stay only where they help
  scanning: the deployment chips.

## intro.html

- Deploy fan-out: five targets that no longer overlap — laptop/coding,
  container/pr-review, edge device/log-triage, cloud vm/autopilot,
  air-gapped host/offline-audit. It previously showed both "edge device" and
  "edge gateway", and "scratch container" next to "k8s pod".
- The caption answers the question the air-gapped node raises: "anywhere you can
  copy a file, against any model you can reach."
- Fix: `fit()` only recomputed on window resize, so an iframe still sized
  300x150 when the script ran kept a 0.21 scale forever. A ResizeObserver on the
  document element fixes it — this was masked while the embed sat inside a dark
  bezel.
- Embedded runs hide the keyboard-hint HUD; those keys need focus the frame does
  not have.

Subpages (getting started, docs, both languages) inherit the system and their
nav links follow the renamed sections. Checked at 1440px and 390px in Chrome,
both languages, no console errors.

### DIFF
--- a/site/index.html
+++ b/site/index.html
@@ -3,10 +3,10 @@
 <head>
 <meta charset="UTF-8">
 <meta name="viewport" content="width=device-width, initial-scale=1.0">
-<title>San — Fast, open agent harness for the terminal</title>
-<meta name="description" content="San is a fast, open terminal agent harness: one ~12 MB Go binary connecting models, tools, permissions, memory, persona bundles, skills, plugins, MCP, and subagents.">
-<meta property="og:title" content="San — Fast, open agent harness for the terminal">
-<meta property="og:description" content="One small Go binary. Bring any model and extension; San gives them a fast, inspectable agent loop in your terminal.">
+<title>San — Minimal overhead. Maximum agent.</title>
+<meta name="description" content="San is an open-source terminal agent runtime: one 12 MB Go binary, ~2.3k tokens of harness, no Node.js or Python. Prompt, tools, providers and extensions all stay yours to change.">
+<meta property="og:title" content="San — Minimal overhead. Maximum agent.">
+<meta property="og:description" content="Small context, native speed, open all the way down. An open-source terminal agent runtime in one 12 MB Go binary.">
 <meta property="og:type" content="website">
 <meta property="og:url" content="https://genai-io.github.io/san/">
 <meta property="og:image" content="https://genai-io.github.io/san/san.png">
@@ -23,13 +23,12 @@
     <span class="logo"><a href="#top">&lt; SAN <span class="spark">✦</span> /&gt;</a></span>
     <button class="nav-toggle" aria-label="Toggle menu" aria-expanded="false" onclick="toggleNav(this)">≡</button>
     <div class="nav-links" id="nav-links">
-      <a href="#features">Features</a>
-      <a href="#deploy">Deploy</a>
-      <a href="#footprint">Footprint</a>
+      <a href="#why">Why</a>
+      <a href="#benchmark">Benchmark</a>
       <a href="#community">Community</a>
       <span class="nav-sep" aria-hidden="true"></span>
       <a href="intro.html">Intro</a>
-      <a href="getting-started.html">Get Started</a>
+      <a href="getting-started.html">Start</a>
       <a href="docs.html">Docs</a>
       <span class="lang-switch">
         <button class="lang-btn" id="lang-btn" aria-haspopup="listbox" aria-expanded="false">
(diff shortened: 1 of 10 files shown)


site/intro.html
```diff
@@ -429,13 +433,16 @@
 
 function fit(){ $('stage').style.transform = `scale(${Math.min(innerWidth/1280, innerHeight/720)})`; }
 addEventListener('resize', fit); fit();
```
hchenxa: **Minor**: The new `ResizeObserver(fit).observe(document.documentElement)` (line 438) already captures viewport changes — the `<html>` element resizes with the viewport. The old `addEventListener('resize', fit)` on line 435 is now redundant, and on window resize both fire back-to-back. Consider removing the `resize` listener since the observer already covers both the window-resize and the iframe-embed cases.

site/index.html
```diff
@@ -38,133 +37,183 @@
...
+    <div class="install load-up d5">
+      <div class="install-tabs" role="tablist" aria-label="Install method">
+        <button class="install-tab active" role="tab" aria-selected="true" data-cmd="curl -fsSL https://raw.githubusercontent.com/genai-io/san/main/install.sh | bash">curl</button>
+        <button class="install-tab" role="tab" aria-selected="false" data-cmd="irm https://raw.githubusercontent.com/genai-io/san/main/install.ps1 | iex">powershell</button>
+        <button class="install-tab" role="tab" aria-selected="false" data-cmd="go install github.com/genai-io/san/cmd/san@latest">go install</button>
+      </div>
+      <div class="install-body">
+        <span class="prompt">$</span>
+        <code id="install-cmd">curl -fsSL https://raw.githubusercontent.com/genai-io/san/main/install.sh | bash</code>
+        <button class="copy-btn" onclick="copyInstall(this)">Copy</button>
```
hchenxa, line 68: **Style consistency (pre-existing, not blocking)**: The page mixes inline `onclick` handlers (`toggleNav`, `copyInstall`) with `addEventListener` calls for install tabs, stage tabs, scroll-spy, and ResizeObserver. Moving the copy handler into the script block would keep event binding in one place. The `toggleNav` pattern is shared across all subpages so that would be a larger refactor — `copyInstall` / `copyEl` would be the easier first step.

hchenxa (commented):
## Review: docs(site): rebuild the homepage around 三

Solid rebuild. The site now reads as one coherent surface — the trigram structure is a clear improvement over the old emoji-card grid, and the design-system reduction (two font families, no glow stacks, matching parchment with intro.html) gives it a distinctive editorial voice that matches the README.

A few observations below.

yanmxa:
Both fixed in this PR.

**`intro.html` resize listener** — right, and it was this branch's regression: the `resize` listener was already on `main`, the `ResizeObserver` is new here. Listener removed. Kept the bare `fit()` call, since the observer's first callback is async and the stage would otherwise render one frame unscaled.

**Inline `onclick`** — done across all six pages, not just the copy handlers. `toggleNav` binds to its own button; the copy buttons name their source with `data-copy`, which collapses `copyInstall` and `copyEl` into one binding per page. No inline handlers left in the site. Verified in the browser: copy on all four pages (including after an install-tab switch), nav toggle open/close with `aria-expanded`, no console errors.

Also, the two `san-ci` comments above are stale — the branch is one signed-off commit and the title follows Conventional Commits.

PR #429, opened 2026-07-26
feat: detect bare image paths on drag-drop to terminal

hsain9357:
When a user drags an image into the terminal, the absolute path is pasted as plain text. This PR:

- Adds `bareImagePathRe` regex matching image paths containing a directory separator
- Extends `ProcessImageRefs` to detect bare paths alongside the existing `@`-prefixed syntax
- Refactors the function into a two-step pipeline for clarity
- Only matches paths with `/` or `\\` to avoid treating bare filenames as images
- Non-existent files are left as-is in the text

### DIFF
--- a/internal/app/input/on_textarea.go
+++ b/internal/app/input/on_textarea.go
@@ -275,52 +275,29 @@ func (m *Model) HistoryDown() {
 // from input. Returns the cleaned text content and any loaded images.
 // Only processes references where the file actually exists on disk;
 // non-existent file references are left in the text as-is.
+//
+// @-prefixed references are removed from the content after loading, and a
+// failed load aborts the entire send — both are right because the user
+// deliberately typed @. Bare paths (drag-drop, absolute paths) are treated
+// more leniently: the text stays in the message, and a failed load is
+// silently skipped so a mere mention of a path doesn't block the send.
 func ProcessImageRefs(cwd, input string) (string, []core.Image, error) {
 	content := input
 	var images []core.Image
 
-	// Step 1: Process @-prefixed image references
-	content, imgList, err := processRefs(cwd, content, imageRefPattern, true)
-	if err != nil {
-		return "", nil, err
-	}
-	images = append(images, imgList...)
-
-	// Step 2: Process bare image file paths (drag-drop, absolute paths, etc.)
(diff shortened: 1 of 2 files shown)


yanmxa:
Thanks @hsain9357 — drag-drop paste is a real gap, and this is a nice first contribution. One thing to adjust; the rest is optional.

Bare paths inherit two behaviors built for `@`: the matched text is removed from the message, and a failed load aborts the send. Both are right when the user typed `@`, but not for a path merely mentioned in a sentence. On this branch:

```
"the golden file docs/assets/golden.png is stale"
  -> "the golden file  is stale"     filename gone from the message
"why does testdata/broken.png fail to decode?"
  -> error, message never sends      (update_submit.go:133)
```

So for bare paths: attach the image but leave the path text in place, and skip a path that fails `image.Load` rather than returning an error. That also clears up the duplicate load when the same path appears twice.

Optional, whenever you like: `processRefs` takes `hasPrefix` but never reads it (both patterns capture in group 1), and a test on `ProcessImageRefs` itself would be a good companion to the pattern test.

Happy to take this once the first part is in.

hsain9357:
Thanks for the thorough review @yanmxa. I've addressed both issues:

1. **Bare paths now keep their text** — the path stays in the message body. Only `@`-prefixed references are stripped from the text.

2. **Failed `image.Load` on bare paths skips silently** — no longer aborts the send. Corrupt or unreadable images are just skipped.

3. Removed the shared `processRefs` helper — `@` and bare path handling now have separate, explicit logic since their semantics differ.

4. Added `TestProcessImageRefs` covering both `@` and bare path behaviors (valid load, corrupt file, missing file, no images).

yanmxa:
Looks right — the two semantics are cleanly separated and the tests cover both.

Correction to my earlier comment: keeping the text doesn't fix the duplicate load. `compare a.png and a.png` still attaches the same file twice. Dedupe by resolved path if you want it; not holding the PR for that.

Could you move the DeepSeek commit (`01098808`) to its own PR? Real gap — `imagesBlockedForModel` only checks the new message, so history images still reach the provider — but it isn't DeepSeek-specific, and we squash-merge, so it would vanish under a drag-drop title.

Drop it and this is good to go. It'll need the same sign-off and signature as #430.

PR #430, opened 2026-07-26
feat: show all files in @ autocomplete and respect .gitignore

hsain9357:
Two changes to the `@` file autocomplete:

1. **Show all files** — removed the `supportedFileExtensions` filter that limited suggestions to only `.md` and image files. Every file type now appears in `@` autocomplete.

2. **Respect .gitignore** — added a gitignore parser (`loadGitignore`) and matcher (`matchGitignorePattern`) supporting basic gitignore patterns: negation (`!`), anchored paths (`/`), directory-only (`/` suffix), and `**` globs. Scanned files and directories are filtered against .gitignore before appearing in suggestions. The hardcoded skip list is kept as a fallback.

### DIFF
--- a/internal/app/kit/suggest/suggest.go
+++ b/internal/app/kit/suggest/suggest.go
@@ -160,15 +160,6 @@ func (s *State) updatefileSuggestions(query string) {
 	s.clampViewStart()
 }
 
-var supportedFileExtensions = map[string]bool{
-	".md":   true,
-	".png":  true,
-	".jpg":  true,
-	".jpeg": true,
-	".gif":  true,
-	".webp": true,
-}
-
 func (s *State) scanAllFiles() []fileSuggestion {
 	seen := make(map[string]bool)
 	var results []fileSuggestion
@@ -179,6 +170,9 @@ func (s *State) scanAllFiles() []fileSuggestion {
 	}
 	queue := []queueItem{{s.cwd, 0}}
 	dirsVisited := 0
+
+	// Load gitignore patterns for filtering
+	gi := loadGitignore(s.cwd)
 
 	for len(queue) > 0 && len(results) < fileScanMaxResults && dirsVisited < fileScanMaxDirsVisited {
 		item := queue[0]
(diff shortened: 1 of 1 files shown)


yanmxa:
Thanks @hsain9357 — `@` should reach any file, and pairing that with `.gitignore` is the right instinct. Net effect is a clear step up: reachable files go from 107 to 500 on this tree. One thing to sort out before merge; the rest is optional.

The 500-result cap. `fileScanMaxResults` is 500 and `scanAllFiles` is cached per cwd, so the first BFS pass becomes the whole `@` index for the session. Measured on this branch against its own tree:

```
depth 0-2   421/421 reachable
depth 3      78/319
depth 4-6      0/29     -> 270 files never offered
```

The cliff also costs 12 files that main does offer today — `.san/personas/software-engineer/skills/*/SKILL.md` and `internal/command/prompts/simplify.md` among them. Raising the cap is enough for now; matching lazily against the query is the nicer fix if you feel like it.

Optional — two cases the matcher currently gets wrong (both measured):

- a lone `**` line ignores every file
- `**/build` also matches `mybuild` and `a/rebuild`

If you'd rather not hand-roll gitignore semantics, happy to discuss pulling in a glob dependency instead.

Nice to have, not required: the package has no tests today, so even a couple of cases here would be a good start.

hsain9357:
Thanks @yanmxa. I've raised the caps:

- `fileScanMaxResults`: 500 → **2000**
- `fileScanMaxDirsVisited`: 2000 → **8000**

That should cover the 270 files at depth 3+ that were hidden. On the `**` matching bugs — I'll leave the hand-rolled matcher as-is for now and can revisit with a proper glob library in a follow-up if needed.

yanmxa (approved):
Cap raise does it — 769/769 files reachable now against 107 on main, ~14ms, once per cwd. That was the blocker.

Two glob bugs for the follow-up, not this PR:

- `**/build` also matches `mybuild`
- `**/*.log` doesn't match `app.log` at all — the part after `**` is compared as a literal suffix, so the `**/*.ext` family never applies

And I was wrong about the lone `**` line: git ignores everything for that rule too, so yours is correct.

Nice work on your first contribution.

yanmxa:
@hsain9357 Really appreciate your work on this!

hsain9357:
> @hsain9357 Really appreciate your work on this!

You are very welcome, your AI Agent fixes a lot of issues I had previously with other AI Agents and it really deserves the contribution.

PR #432, opened 2026-07-27
fix: strip images from messages sent to DeepSeek

hsain9357:
DeepSeek's Chat Completions API rejects `image_url` content parts. When conversation history contains images — from bare-path detection or a prior session with a vision-capable model — those images get serialized as `image_url` and the request fails.

Strip Images from all messages before sending to DeepSeek. This is a text-only provider fix; vision-capable providers (Gemini, Anthropic, OpenAI) are unaffected.

### DIFF
--- a/internal/app/agent.go
+++ b/internal/app/agent.go
@@ -452,7 +452,26 @@ func (m *model) seedAgentMessages(pendingSend string) []core.Message {
 			coreMessages = coreMessages[:len(coreMessages)-1]
 		}
 	}
-	return coreMessages
+	return m.dropImagesTextOnlyModelRejects(coreMessages)
+}
+
+// dropImagesTextOnlyModelRejects removes image attachments from a seeded chain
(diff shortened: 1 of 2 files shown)


yanmxa:
Good catch, and the gap is real — I reproduced it on main. Image in the history, switch to DeepSeek, send plain text: the payload still carries `image_url`, because `imagesBlockedForModel` only ever sees the new message and the queued item, never the history.

One ask: could this move up to the app layer, next to `imagesBlockedForModel`? It already knows the answer via `llm.SupportsImages` — it just misses the history. Fixing it there covers every provider, including the ones that don't go through `openaicompat`. (Correcting myself from #429: DeepSeek is in fact the only provider declaring `SupportsImages() == false` today, so this isn't duplicated anywhere yet — it's about where the gap actually lives.)

Also needs a rebase — it branched off #429, which is now merged, so those two commits are still in here. And the commits need a sign-off.

PR #437, opened 2026-07-31
fix: keep assistant text visible while a docked modal is up

### LINKED ISSUE #436: Assistant text before a tool call is hidden while the permission modal is open
## Problem

When the agent emits reasoning text and then immediately requests permission for a tool call, the approval modal hides the text that was just streamed. Only after the user approves (or rejects) does the text become visible in the conversation.

Example flow:

1. Assistant streams:
   > A key clue emerged: `dial tcp ...: connect: invalid argument` — this is not a regular timeout but a kernel-level network connectivity issue. Let me dig into the kernel networking state:
2. Assistant then requests permission: `● Bash  Check node conditions`
3. **While the approval dialog is open, the text from step 1 is not visible** — the user sees only the bare tool request, without the context explaining *why* this command is being run (in this case, that the investigation is network-related).
4. After approving, the text from step 1 appears.

This makes it hard to make an informed decision on the permission request, because the agent's rationale immediately preceding the request is hidden at exactly the moment the user needs it.

## Expected behavior

While the permission approval modal is displayed, the assistant text that precedes the tool call should remain visible above the modal, so the user can read the agent's reasoning and understand what they are approving.

## Likely cause (from a quick code read)

In `internal/app/view.go`, `viewString()` short-circuits for docked modals (Question / Approval) and renders only `trackerPrefix + separator + ov.Render()` — the active chat content (`RenderActiveContent`, which holds the uncommitted assistant text) is skipped. The text only re-enters the view once the modal closes and the normal layout (`renderNormalView`) resumes.

yanmxa:
The docked-modal branch of `viewString` rendered only the modal, so the assistant text streamed right before a tool call — the rationale for the permission being requested — was hidden until the modal closed. Fixes #436.

- The live chat tail now renders above the modal, capped to the rows the modal leaves free, so a long explanation can never push the answer options off screen.
- Rendering the tail exposed a second problem: `PreToolEvent` stamps a call as started *before* the permission request, so its row spun with a ticking timer while the user was still deciding. `RenderContext.DockedModalActive` (the widened, renamed successor to `InteractivePromptActive`) now holds every in-flight row still until the modal is answered — a parallel batch is stuck behind the same answer.
- Both layouts budget their height through one helper, `chatTailAbove`, which also skips the render outright when the bottom block already fills the screen.

Manually verified in a real session: approve, Esc-reject, a 60-sentence explanation (frame stays exactly at terminal height, truncated from the top), a parallel Bash batch, and the AskUserQuestion modal — which had the same hidden-text bug and is fixed by the same change.

### DIFF
--- a/internal/app/conv/message.go
+++ b/internal/app/conv/message.go
@@ -194,6 +194,10 @@ type AssistantParams struct {
 	ToolCallsExpanded bool
 	StreamActive      bool
 	IsLast            bool
+	// DockedModalActive freezes the bullet: the stream stays "active" across a
+	// permission gate (it clears only on a text-only Done chunk), so without
+	// this the live tail spins while the turn is parked on the user's answer.
+	DockedModalActive bool
 	SpinnerView       string
 	MDRenderer        *MDRenderer
 	Width             int
@@ -338,7 +342,7 @@ func RenderAssistantMessage(params AssistantParams) string {
 	// committed the bullet is spent and the remainder aligns under a gutter.
 	// While streaming, the active tail shows the spinner in the bullet slot.
 	aiIcon := contentGutter(!params.BulletEmitted)
-	if params.StreamActive && params.IsLast {
+	if params.StreamActive && params.IsLast && !params.DockedModalActive {
 		aiIcon = aiPromptStyle.Render(params.SpinnerView + " ")
 	}
 
@@ -510,6 +514,15 @@ type ToolCallsParams struct {
 	Interactive  bool
 }
 
+// expandHints reports whether "(ctrl+o to expand)" affordances belong on this
+// pass. They ride the live tail only, and not while a docked modal owns the
+// keyboard: routeKeypress hands every key to the active overlay, so ctrl+o
+// reaches the modal (where it toggles a preview, or nothing) and never the row
+// offering the hint.
+func (p ToolCallsParams) expandHints() bool {
+	return p.Interactive && !p.DockedModalActive
+}
+
 // ToolResultData holds the data needed to render a tool result inline.
 type ToolResultData struct {
 	ToolName    string
@@ -552,8 +565,15 @@ func RenderToolCalls(params ToolCallsParams) string {
 			if hasResult {
 				sb.WriteString(renderAgentToolLine(label, params.Width, "●", color) + "\n")
 			} else {
-				sb.WriteString(renderAgentToolLine(label, params.Width, agentIcon(params.Blink), color))
-				if !params.ToolCallsExpanded && params.Interactive {
+				// agentIcon blinks ●/○ off the frame counter; under a docked
+				// modal it would advertise a subagent as working while the
+				// turn waits on the user, same as a spinning tool row.
+				icon := agentIcon(params.Blink)
+				if params.DockedModalActive {
+					icon = "●"
+				}
+				sb.WriteString(renderAgentToolLine(label, params.Width, icon, color))
+				if !params.ToolCallsExpanded && params.expandHints() {
 					sb.WriteString(ThinkingStyle.Render("  (ctrl+o to expand)"))
 				}
 				sb.WriteString("\n")
@@ -579,9 +599,6 @@ func RenderToolCalls(params ToolCallsParams) string {
 			}
 		} else {
 			icon := toolCallIcon(tc, params)
-			if _, hasResult := params.ResultMap[tc.ID]; hasResult {
-				icon = "●"
-			}
 			detail := runningRowDetail(tc, params)
 			var row string
 			if tc.Name == tool.ToolTaskGet && params.TaskOwnerMap != nil {
@@ -599,7 +616,7 @@ func RenderToolCalls(params ToolCallsParams) string {
 
 		if resultData, ok := params.ResultMap[tc.ID]; ok {
 			resultData.ToolInput = tc.Input
-			resultData.Interactive = params.Interactive
+			resultData.Interactive = params.expandHints()
 			resultData.Width = params.Width
 			resultData.Nested = true
 			// Decision sits between the call and its result, mirroring the
@@ -623,12 +640,22 @@ func RenderToolCalls(params ToolCallsParams) string {
 }
 
 // toolCallIcon picks a call's row glyph: the spinner while it is in flight, a
-// static bullet otherwise. Under a docked modal every row is static — the turn
-// is parked on the user's answer. The call the modal is asking about was marked
+// static bullet once it has a result or is not the call being executed.
+//
+// A docked modal freezes every row. The call it is asking about was marked
 // current and stamped as started by its PreToolEvent, which fires *before* the
 // permission request, so it would otherwise spin as if it had been allowed to
-// run; siblings in the same batch are just as stuck behind the same answer.
+// run. Nothing here can tell that call apart from a batch sibling that really
+// is executing (the permission gate lives inside each tool's Execute, and a
+// parallel batch runs one goroutine per call), so the freeze is deliberately
+// blunt: a row that stalls for the length of the decision beats a row that
+// claims a command is running before it was allowed to start. Distinguishing
+// them needs the pending call's identity on the permission request itself —
+// see #440.
 func toolCallIcon(tc core.ToolCall, params ToolCallsParams) string {
+	if _, done := params.ResultMap[tc.ID]; done {
+		return "●"
+	}
 	if params.DockedModalActive {
 		return "●"
 	}
--- a/internal/app/conv/message.go
+++ b/internal/app/conv/message.go
@@ -486,6 +486,9 @@ type ToolCallsParams struct {
 	ToolCallsExpanded bool
 	ResultMap         map[string]ToolResultData
 	ParallelMode      bool
+	// DockedModalActive marks the frames where a Question / Approval modal
+	// owns the screen; the call it is asking about holds still until answered.
+	DockedModalActive bool
 	TaskActivity      map[int][]string
 	PendingCalls      []core.ToolCall
 	CurrentIdx        int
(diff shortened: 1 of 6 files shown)


internal/app/conv/view.go
```diff
@@ -256,6 +260,7 @@ func renderAssistantWithTools(p RenderContext, msg core.ChatMessage, idx int, is
 		ToolCallsExpanded: msg.ToolCallsExpanded,
 		ResultMap:         resultMap,
 		ParallelMode:      len(p.PendingCalls) > 1,
+		DockedModalActive: p.DockedModalActive,
```
hchenxa, line 264: The freeze stops here: `DockedModalActive` is threaded into `ToolCallsParams` and `PendingToolSpinnerParams`, but never into `AssistantParams`. `RenderAssistantMessage` (message.go:340-343) still paints `params.SpinnerView` in the leading bullet whenever `StreamActive && IsLast` — and `Stream.Active` stays true while the turn is parked on the permission gate (it is only cleared on a text-only Done chunk, conv/update.go:211-212). So the rationale row above the modal carries a rotating spinner for the entire decision — the same "spinning while parked" lie this PR fixes for tool rows. Reproduce: any modal over the last uncommitted assistant message; its bullet rotates at tick cadence. The new test fixture cannot catch this because it never sets `Stream.Active`.

internal/app/conv/message.go
```diff
@@ -575,7 +578,7 @@ func RenderToolCalls(params ToolCallsParams) string {
 				}
 			}
 		} else {
-			icon := toolCallIcon(tc, params.PendingCalls, params.CurrentIdx, params.ParallelMode, params.SpinnerView)
+			icon := toolCallIcon(tc, params)
```
hchenxa, line 601: This call now gates the *standard* tool rows, but the agent-tool branch above (line 555) renders `renderAgentToolLine(label, params.Width, agentIcon(params.Blink), color)` with no `DockedModalActive` check — `agentIcon` alternates ●/○ every ~720ms while the ticker runs (`needsSpinner()` holds while `Stream.Active`). A `SpawnAgent` call prompts for permission by default, so the modal goes up over its own row, which blinks "working" the whole time the user decides. `RenderPendingToolSpinner`'s agent path is gated (agent_activity.go:225); this inline one was missed.

hchenxa (commented):
Reviewed the fix end-to-end (ran the new tests — they pass). Overall this is a clean, well-tested fix: the height budget arithmetic closes at exactly terminal height, the `renderNormalView` refactor is output-identical, and the old `InteractivePromptActive` flag was indeed dead code. Comments below cover the freeze's gaps (findings 1–3), one efficiency point (4), and a few smaller nits (5–7).

One thing I checked but did not comment inline: `renderAndCommit` (model_scrollback.go:241) can run while a Question modal is up and would commit a frozen spinner glyph — but that is pre-existing behavior, byte-identical before this PR, so I left it out of scope.

internal/app/conv/message.go
```diff
@@ -619,9 +622,19 @@ func RenderToolCalls(params ToolCallsParams) string {
...
 
-func toolCallIcon(tc core.ToolCall, pendingCalls []core.ToolCall, currentIdx int, parallelMode bool, spinnerView string) string {
+// toolCallIcon picks a call's row glyph: the spinner while it is in flight, a
+// static bullet otherwise. Under a docked modal every row is static — the turn
+// is parked on the user's answer. The call the modal is asking about was marked
+// current and stamped as started by its PreToolEvent, which fires *before* the
+// permission request, so it would otherwise spin as if it had been allowed to
+// run; siblings in the same batch are just as stuck behind the same answer.
+func toolCallIcon(tc core.ToolCall, params ToolCallsParams) string {
+	if params.DockedModalActive {
```
hchenxa, line 659: The freeze is over-broad: it also freezes siblings that are *genuinely executing*. Permission is gated per call inside each tool's `Execute` (internal/tool/permission.go:51-56), and parallel batches run per-call goroutines (core/agent_impl.go:535-538 — "Agent calls are independent workers with their own permission gates — running them concurrently is the point"). So in a parallel batch [A allow-listed, B prompts], A runs for real while B's modal is up, but A's row shows a static ● with no elapsed time; when B is answered, A's result and timer materialize at once. The comment above ("siblings in the same batch are just as stuck behind the same answer") is only true when every call in the batch is gated. Stamping `MarkStarted` at actual execution start (on approval) instead of at PreTool would make the row honest without needing a freeze flag.

internal/app/view.go
```diff
@@ -87,21 +103,32 @@ func isDockedModal(ov overlayPanel) bool {
...
+// its earliest lines are the ones that scroll off.
+func (m *model) chatTailAbove(bottom string, params conv.RenderContext, trackerView string) string {
+	// A non-positive budget means the bottom block already fills the screen.
+	// Bail before rendering: whatever came back would be dropped whole, and
+	// letting it through would push the block off the bottom of the frame.
+	maxContentHeight := m.env.Height - strings.Count(bottom, "\n")
+	if maxContentHeight <= 0 {
+		return ""
+	}
+	return tailLines(m.renderChatSection(conv.RenderActiveContent(params), trackerView), maxContentHeight)
```
hchenxa, line 129: Every tick (the spinner cadence, ~360ms — `needsSpinner()` holds while `Stream.Active`) the entire uncommitted tail is rendered — markdown, tool rows, spinner, tracker — and only then trimmed to the budget; before this PR the modal branch rendered no tail at all. With a long rationale (the PR's own test uses a ~2.8KB wall) most of the render work is discarded every frame. The glamour cache makes repeat markdown passes cheap, but block assembly, tool rows, and the split/join trim still run on the full content. Consider capping the raw lines to the budget before the heavy render.

internal/app/view.go
```diff
@@ -55,15 +55,31 @@ func (m *model) viewString() (string, *tea.Cursor) {
...
+// The modal claims its rows first and the chat tail gets what is left: rendering
+// the tail unbounded would push the modal's answer options off the bottom of the
+// screen, which is worse than showing no text at all.
+func (m *model) renderDockedModalView(ov overlayPanel, separator, trackerView string) string {
+	modal := "\n" + separator + "\n" + ov.Render()
+
+	params := m.messageRenderParams()
+	// Only this branch renders under a modal, so it is the one place that knows
+	// the turn is parked on an answer — everything animated holds still.
+	params.DockedModalActive = true
```
hchenxa, line 78: The tail rendered here runs `RenderActiveContent`, which sets `Interactive = true` — so pending agent rows print "(ctrl+o to expand)" (message.go:556-558) above the modal. But routeKeypress hands ctrl+o to the active overlay first (update_keys.go:36-38), and the Approval modal's ctrl+o only toggles diff/bash previews — a no-op for agent approvals. The hint is a dead affordance exactly when the modal is up; consider suppressing the `Interactive` hints on this path.

internal/app/view.go
```diff
@@ -87,21 +103,32 @@ func isDockedModal(ov overlayPanel) bool {
...
+// block leaves free — the footer in the normal layout, the modal in the docked
+// one. Both layouts anchor to the bottom of the screen, so this is the single
+// place the height budget is decided: the tail is capped to what is left, and
+// its earliest lines are the ones that scroll off.
+func (m *model) chatTailAbove(bottom string, params conv.RenderContext, trackerView string) string {
+	// A non-positive budget means the bottom block already fills the screen.
+	// Bail before rendering: whatever came back would be dropped whole, and
+	// letting it through would push the block off the bottom of the frame.
+	maxContentHeight := m.env.Height - strings.Count(bottom, "\n")
+	if maxContentHeight <= 0 {
```
hchenxa, line 126: Two nits: (1) this `maxContentHeight <= 0` bail duplicates the identical guard in `tailLines` (view.go:152-154) — output is identical without it (it only skips the render work), and TestTailLines pins the `<= 0 → ""` contract. (2) The comment is off by one: "non-positive budget means the bottom block already fills the screen" — an exact fill (Height rows) yields budget 1, not non-positive; non-positive means the block is strictly *taller* than the screen. And "letting it through would push the block off the bottom of the frame" is inaccurate — `tailLines` drops it, nothing is pushed.

internal/app/conv/message.go
```diff
@@ -575,7 +578,7 @@ func RenderToolCalls(params ToolCallsParams) string {
 				}
 			}
 		} else {
-			icon := toolCallIcon(tc, params.PendingCalls, params.CurrentIdx, params.ParallelMode, params.SpinnerView)
+			icon := toolCallIcon(tc, params)
 			if _, hasResult := params.ResultMap[tc.ID]; hasResult {
```
hchenxa: Since `toolCallIcon` now receives the full params, the "row with a result shows ●" decision is split across two places — this override and the spinner logic inside `toolCallIcon`. Folding `if _, done := params.ResultMap[tc.ID]; done { return "●" }` into `toolCallIcon` (mirroring `runningRowDetail`'s first check, line 664) is exactly equivalent on every path I traced (pending-with-result in a parallel batch — MarkComplete keeps completed calls in PendingCalls until the batch finishes; sequential non-current; idx == -1; and under the modal) and makes the icon a single source of truth.

yanmxa:
Thanks — all confirmed. Fixed in 42cfd4a7:

- **1** assistant bullet: the flag now reaches `AssistantParams`; the fixture sets `Stream.Active`, so it is covered.
- **2** agent rows: the `IsAgentToolName` branch bypassed `toolCallIcon` entirely — frozen, tested at both blink phases.
- **5** ctrl+o hint: both sites go through `expandHints()` (`Interactive && !DockedModalActive`).
- **6(2)** comment rewritten — an exact fill is budget 1, and `tailLines` drops the tail.
- **7** result-map check folded into `toolCallIcon`.

**3** — right that "siblings are just as stuck" is false. The render path still cannot tell the gated call from a running one, so the freeze stays blunt and the comment now states that as a trade-off. The real fix (identity on the permission request, `MarkStarted` on approval) is #440, including the `renderAndCommit` case you flagged.

Keeping two: **4** — line count only exists after wrapping, so there is nothing to cap before the render; **6(1)** — the bail exists to skip that render, which is entirely wasted on a terminal shorter than the modal.

Verified manually: bullet and tool rows hold still across frames under the modal, and animate again once it is answered.

PR #438, opened 2026-07-31
feat: inline image paths for text-only models instead of blocking

hsain9357:
## Summary

When the active model is text-only (e.g. DeepSeek), a turn carrying image attachments was previously held back with the notice *"The current model doesn't support images — remove the image or switch to a vision-capable model."*

Instead of blocking, the image's **absolute path** is now inlined into the message content and the binary attachment is dropped, letting the model itself decide what to do — e.g. call an MCP image-description tool with the path (see the explain-image style servers) — rather than the app refusing the turn.

Vision-capable models are unaffected and still receive the binary attachments.

## Changes

- `core.Image` gains a `Path` field (absolute source path; empty for clipboard pastes).
- `image.Load` records the absolute path; new `image.ResolvePath` returns it, or materializes clipboard data to a temp file so text-only models always get a usable path.
- Replaces `imagesBlockedForModel` (notice + hold-back) with `adaptImagesForModel`, applied on the direct submit, input-queue drain, and turn-boundary release paths.
- Removes the blocking notice entirely.

## Test plan

- `CGO_ENABLED=0 go build ./...` / `go vet ./...` — clean
- `CGO_ENABLED=0 go test ./...` — all 67 packages pass

### DIFF
--- a/internal/app/model_turn_queue.go
+++ b/internal/app/model_turn_queue.go
@@ -48,8 +48,8 @@ func (m *model) handleStopHookResult(msg stopHookResultMsg) tea.Cmd {
 // releaseQueuedMessage hands the next queued user message to the running agent
 // and returns (cmd, true) when one was released, or (nil, false) when the queue
 // is empty or its head is under edit (SelectIdx 0 — dispatching would send
-// pre-edit text and orphan the user's changes). An image-blocked item is diverted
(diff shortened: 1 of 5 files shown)


hsain9357:
sorry i've already made a PR on this issue and it was fixed. I've uploaded this one by mistake.

PR #439, opened 2026-07-31
feat: hand text-only models the image path instead of blocking the message

hsain9357:
## Summary

Two things a user hits when they put an image into a prompt while a text-only model is selected.

**A message carrying an image was blocked outright.** With DeepSeek — or any provider declaring `SupportsImages() == false` — the turn was held back behind *"The current model doesn't support images — remove the image or switch to a vision-capable model."* Now each image's absolute path is inlined into the content and the attachment is withheld from the provider, so the model can decide what to do with the path (call an MCP image-description tool on it, say) instead of the app refusing to send. Vision-capable models are untouched and still receive the binary attachments. No image URL is ever produced — local filesystem paths only.

**A prompt starting with an absolute image path parsed as a slash command.** Dragging an image in before typing produces `/Users/me/shot.png what is this`, whose first token begins with `/`, so it came back as `Unknown command: /Users/me/shot.png` and never reached the agent. Slash-command matching is now skipped when the first token is a path to an existing image, and that path is consumed as an image reference rather than left in the text.

## What the review rounds changed

**Images still reached the provider.** The first cut inlined the paths but handed the attachments to `Send` anyway, on the assumption that `dropImagesTextOnlyModelRejects` would strip them downstream. It doesn't: that helper runs only inside `seedAgentMessages`, which drops the pending message from the chain and is skipped altogether once the session is active. So every turn with an image still produced the `image_url` part DeepSeek rejects — the request #432 fixed — and it stayed in the agent chain, replaying on every later turn. `adaptTurnForProvider` now returns the images the *provider* may receive (none, for a text-only model) while the message appended to `conv` keeps its own, so the transcript still shows the image and a later switch to a vision-capable model can use it.

**Two more chains go straight from `conv` to the active model** and needed the same treatment: the `/compact` summarization request, and the next-input hint. Before this branch they only misfired for a conversation carried over from a vision model; now that `conv` keeps images on a text-only model, a `/compact` after sending one image would 400.

**A failed queued image took the message being typed with it.** `releaseQueuedMessage` called `userInput.Reset()` on an image-load error, which runs mid-stream — exactly when the textarea holds the next message. It also resolved `[Image #N]` markers against the textarea's *current* pending images rather than the queued item's own, so an image pasted during the turn was picked up by the queued message. Both fixed, and the queued release now makes the same display/content split `buildUserMessage` makes, so a queued message renders like a directly submitted one.

**`env.ProjectRoot` was dropped.** It was assigned once at startup and never again, while `CWD` moves when a Bash `cd` fires `changeCwd`. Resolving image references against the frozen copy put it at odds with the `@`-completion popup, which lists files from the live cwd — after a `cd`, the popup offers a file `ProcessImageRefs` then can't find. Both follow `CWD`.

## Changes

- `internal/app/update_submit.go` — `adaptTurnForProvider`, leading-image-path bypass
- `internal/app/model_turn_queue.go` — queued release: the item's own images, no textarea reset, display/content split
- `internal/app/model_compact.go`, `internal/app/input/prompt_suggestion.go` — the other two provider-bound chains
- `internal/app/input/on_textarea.go` — `LeadingImagePath`, leading-path consumption
- `internal/core/message.go`, `internal/image/image.go`, `internal/image/clipboard.go` — `Image.Path`, `EnsureFilePath`
- `internal/app/model.go`, `internal/app/model_agent_events.go` — temp files for clipboard images, removed at turn end
- `internal/llm/deepseek/client.go` — `low` / `medium` thinking efforts

## Test plan

- `CGO_ENABLED=0 go build ./...`, `make lint` (vet + layercheck), `gofmt -l` — clean
- `CGO_ENABLED=0 go test ./...` — all packages pass
- New `internal/app/text_only_images_test.go` (6 tests) and `internal/app/input/prompt_suggestion_test.go`: attachments withheld on both send paths, the compact and hint chains, the queued-image error, and the display/content split. Each was checked against the commit before its fix and fails there.
- `internal/app/input/on_textarea_test.go` — 13 table cases for `LeadingImagePath` and leading-path handling in `ProcessImageRefs`

## DeepSeek thinking efforts

The effort list is now `off / low / high / xhigh / max`, matching the Requested-effort column of the [thinking-mode docs](https://api-docs.deepseek.com/guides/thinking_mode/). `medium` was in an earlier revision of this branch and is not a value the API takes — it is forwarded verbatim, so picking it would have 400'd every turn. `off` is a san-side value: `Stream` omits `reasoning_effort` entirely for it.


hsain9357:
Hi team a feature in this PR seem to be in this PR https://github.com/genai-io/san/pull/432 but i have pulled the main branch locally then built it and found the Deepseek stripping of images did not work idk why,  so I've made this PR, please check it and merge it after validation.

hchenxa (commented):
Reviewed all three commits end-to-end (build + tests pass). The design is solid: no send path reaches a text-only model with raw image parts (all `SubmitToAgent` image carriers go through `adaptImagesForModel`, and `dropImagesTextOnlyModelRejects` is a second line of defense at the agent layer), and the leading-path byte-slice strip is safe. Comments below cover one regression from the ProjectRoot change (1), a temp-file leak (2), a coverage gap in the queue path (3), the DeepSeek default flip (4), silent image-dropping (5), a load-failure edge (6), a stat-error nit (7), and dead code left behind (8).

Two notes I did not comment inline: `notes/active/changes-summary.md` documents already-merged PRs (#429–432) and will go stale — consider dropping it from this PR; and `extForMediaType` returns a nondeterministic extension for `image/jpeg` (both `.jpg` and `.jpeg` map to it — latent only, clipboard images are always PNG).

internal/app/model_workspace.go
```diff
@@ -17,7 +17,7 @@ func (m *model) changeCwd(newCwd string) {
 	oldCwd := m.env.CWD
 	m.env.CWD = newCwd
 	m.env.IsGit = m.services.Setting.IsGitRepo(newCwd)
-	m.userInput.HandleCwdChange(newCwd)
+	m.userInput.HandleCwdChange(m.env.ProjectRoot)
```
hchenxa: The picker and the path resolver now disagree about the root. Suggestion entries are **relative** paths (`filepath.Rel(s.cwd, …)`, internal/app/kit/suggest/suggest.go:200-227) — after this change, `s.cwd` is `ProjectRoot` — so the user picks `@sub/photo.png` meaning *ProjectRoot*/sub/photo.png. But `ProcessImageRefs(m.env.CWD, raw)` (internal/app/update_submit.go:139) and `LeadingImagePath(m.env.CWD, raw)` still resolve relative paths against the **runtime** CWD, which `changeCwd` set to wherever the model `cd`'d via Bash (internal/app/model_tool_effects.go:25). After any `cd`, an image picked from the picker resolves against the wrong directory → `os.Stat` ENOENT → silently skipped: no attachment, and the literal `@sub/photo.png` stays in the text the model sees. Before this PR both used the same root, so this is a regression. Fix: resolve `@`/bare relative paths against `ProjectRoot` as well, or insert absolute paths into the textarea on pick.

internal/image/image.go
```diff
@@ -67,15 +67,51 @@ func Load(path string) (core.Image, error) {
...
+		return img.Path, nil
+	}
+	if img.Data == "" {
+		return img.FileName, nil
+	}
+	data, err := base64.StdEncoding.DecodeString(img.Data)
+	if err != nil {
+		return "", fmt.Errorf("decoding image data: %w", err)
+	}
+	f, err := os.CreateTemp("", "san-image-*"+extForMediaType(img.MediaType))
```
hchenxa, line 101: Temp file is never removed — one leaked file per clipboard-image send to a text-only model. `adaptImagesForModel` (internal/app/update_submit.go:173) materializes pasted images here, and nothing in the codebase ever deletes the `san-image-*` file; the repo's own convention is `defer os.Remove` (internal/image/clipboard.go). Files accumulate in TMPDIR across sessions (up to ~5MB each), and the temp path is embedded in the persisted transcript content, so saved conversations permanently reference these files. The caller gets only a bare path with no lifecycle hook — consider returning a cleanup closer, or removing the file when the turn ends.

internal/app/model_turn_queue.go
```diff
@@ -64,14 +64,11 @@ func (m *model) releaseQueuedMessage() (tea.Cmd, bool) {
...
-		// Hand the message back instead of dropping it — the notice tells the
-		// user to remove the image or switch models.
-		m.userInput.ReturnToTextarea(item.Content, item.Images)
-		return tea.Batch(m.CommitMessages()...), true
-	}
-	m.conv.Append(core.ChatMessage{Role: core.RoleUser, Content: item.Content, Images: item.Images})
-	svc, content, images := m.services.Agent, item.Content, item.Images
+	// Text-only models can't receive image parts; inline each image's path so
+	// the model can decide how to use it (e.g. via an MCP tool).
+	content, images := m.adaptImagesForModel(item.Content, item.Images)
```
hchenxa: The leading-image-path fix doesn't cover the queue path. `releaseQueuedMessage` (turn-boundary and step-boundary drains) runs `adaptImagesForModel` but never `ProcessImageRefs` — so a drag-drop image pasted while a turn is streaming is queued as raw text (`enqueueWhileStreaming` stores `raw`) and released with the leading path un-consumed and unloaded: no image is attached, and the raw path reaches the model as plain text. The PR description claims the fix applies on "turn-boundary release paths" — true for path-inlining, not for leading-path consumption. Same applies to `@`-references in queued content.

internal/llm/deepseek/client.go
```diff
@@ -46,14 +46,14 @@ func (c *Client) ThinkingEfforts(model string) []string {
...
-	return []string{"off", "high", "max"}
+	return []string{"off", "low", "medium", "high", "max"}
 }
 
 func (c *Client) DefaultThinkingEffort(model string) string {
 	if !supportsThinking(model) {
 		return ""
 	}
-	return "off"
+	return "medium"
```
hchenxa: Default thinking effort flips `off` → `medium` for every DeepSeek user. `"off"` suppressed the parameter entirely (see the `opts.ThinkingEffort != "off"` guard in this file), so before this PR no `reasoning_effort` was ever sent unless the user opted in; now every request sends `reasoning_effort=medium` — extra reasoning output, cost, and latency on upgrade with no opt-in. Worth (a) calling the default flip out in the PR description, and (b) confirming `low`/`medium` are accepted values for the v4 API since they're sent verbatim (if unknown, the provider maps them rather than erroring).

internal/app/update_submit.go
```diff
@@ -99,19 +100,24 @@ func (m *model) dispatchSubmission(raw string) tea.Cmd {
...
 	}
-	// Hold the turn (keeping the input) if it carries images the active model
-	// can't accept, rather than letting the provider reject the request.
-	if m.imagesBlockedForModel(msg.Images) {
-		return tea.Batch(m.CommitMessages()...)
-	}
+	// Text-only models can't receive image parts, so inline each image's path
+	// into the content and let the model decide how to use it (e.g. call an
+	// MCP tool to inspect the file) instead of refusing the turn.
+	msg.Content, msg.Images = m.adaptImagesForModel(msg.Content, msg.Images)
```
hchenxa: Attached images are silently dropped from the transcript and the conversation for text-only models. `adaptImagesForModel` returns `nil` images, and that's what gets appended and persisted: the user's turn renders with no image thumbnail, and if the user switches to a vision-capable model later in the session (or resumes), the original attachments are gone — only the prose note and path (for clipboard images, a temp-file path) remain. Blocking the turn (the old behavior) preserved the image; worth considering keeping the images on the message while sending the path-inlined content (the agent layer already strips them for text-only providers).

internal/app/input/on_textarea.go
```diff
@@ -328,6 +363,17 @@ func ProcessImageRefs(cwd, input string) (string, []core.Image, error) {
 			continue
 		}
 		images = append(images, img)
+		if !leadingLoaded && LeadingImagePath(cwd, content) == path {
```
hchenxa: A leading image that exists but fails to load is sent as plain text with zero feedback. A drag-drop of an oversized (>maxImageSize) or corrupt file passes the `LeadingImagePath` existence gate (good — no unknown-command error), but `image.Load` fails and the bare path is skipped silently here: the image never attaches and no notice appears, while the doc comment claims the leading path is "consumed as an image reference, mirroring @-prefixed behavior" — which aborts loudly with an error. The user believes the image went with the message. Consider a notice on this path.

Minor: `LeadingImagePath(cwd, content)` is re-evaluated once per loaded match plus once more below — ~N duplicate parse+stat calls on the same first token; compute it once before the loop.

internal/app/input/on_textarea.go
```diff
@@ -271,6 +271,36 @@ func (m *Model) HistoryDown() {
...
+	// comma must disqualify it.
+	m := bareImagePathRe.FindStringSubmatch(first)
+	if m == nil || m[1] != first {
+		return ""
+	}
+	absPath := first
+	if !filepath.IsAbs(absPath) {
+		absPath = filepath.Join(cwd, absPath)
+	}
+	if _, err := os.Stat(absPath); os.IsNotExist(err) {
```
hchenxa: Guard is inverted in intent: `os.IsNotExist(err)` falls through for any *other* stat error (EACCES, ENOTDIR, ELOOP), so `return first` treats a token whose path can't be stat'd as an existing image path — bypassing slash-command matching in `dispatchSubmission` for a token that isn't a usable image. Should be `if err != nil { return "" }`. Minor in practice (rare error conditions), but the two error branches are semantically opposite.

internal/app/update_submit.go
```diff
@@ -148,16 +154,29 @@ func (m *model) buildUserMessage(raw string) (core.ChatMessage, bool) {
...
-// because the active model is text-only. It adds a user-facing notice when it
-// blocks; callers must not send the turn and should preserve the input so the
-// user can remove the image or switch to a vision-capable model.
-func (m *model) imagesBlockedForModel(images []core.Image) bool {
+// adaptImagesForModel adapts a user turn to the active model's image
+// capability. Vision-capable models receive the binary attachments unchanged.
+// Text-only models cannot receive image parts, so each image's path is inlined
+// into the content instead — the model then decides how to use it (e.g. call
+// an MCP image-description tool) rather than the app refusing the turn.
+func (m *model) adaptImagesForModel(content string, images []core.Image) (string, []core.Image) {
```
hchenxa: Dead code left behind by this change: `Model.ReturnToTextarea` (internal/app/input/on_textarea.go:436) had exactly two callers — the `imagesBlockedForModel` branches in `releaseQueuedMessage` and `drainInputQueueWhileIdle` — both removed here; grep finds zero remaining call sites (tests included). Also the comment on `dropImagesTextOnlyModelRejects` (internal/app/agent.go:458-460) still names the deleted `imagesBlockedForModel`. Delete the method and fix the stale reference.

hsain9357:
Thanks for the thorough review @hchenxa. All 8 points have been addressed in the latest push:

1. **ProjectRoot vs CWD** — `LeadingImagePath` and `ProcessImageRefs` now use `m.env.ProjectRoot` instead of `m.env.CWD`, matching the file picker anchor.

2. **Temp-file leak** — clipboard materializations from `ResolvePath` are tracked on `model.tempImageFiles` and cleaned up at `OnTurnEnd`.

3. **Queue-path coverage** — `releaseQueuedMessage` now runs `ProcessImageRefs` on queued content before sending, covering leading-image-path and @-reference gaps.

4. **DeepSeek default** — `DefaultThinkingEffort` reverted from `"medium"` to `"off"`; no silent reasoning_effort on upgrade.

5. **Silent image dropping** — `adaptImagesForModel` returns the original images slice (not nil), so they survive in the transcript and on model switch. The agent layer (`dropImagesTextOnlyModelRejects`) still strips them before sending.

6. **Load-failure edge** — when the leading image passes stat but fails to load (corrupt/oversized), the path is consumed from text and an error is returned for a user-facing notice.

7. **Stat-error guard** — `os.IsNotExist(err)` → `err != nil` so EACCES/ENOTDIR/ELOOP also bypass slash-command matching.

8. **Dead code** — `ReturnToTextarea` deleted (zero callers), stale `imagesBlockedForModel` comment in agent.go updated. Also dropped `notes/active/changes-summary.md` per your note.

Also fixed the import ordering that was flagged by the format check CI step.

hchenxa:
@yanmxa PTAL

yanmxa (commented):
Thanks for turning the eight points around so quickly — reviewed at `b00d42e5`, build and tests pass, and they all look addressed. Two things came in with the fix commit though.

**Images still reach text-only providers.** `adaptImagesForModel` hands the images back and leaves the stripping to the agent layer, but `dropImagesTextOnlyModelRejects` only runs inside `seedAgentMessages` — a normal turn skips it, since `ensureAgentSession` returns early when the session is already active. So the images go straight to `Session.Send`, `openaicompat` makes them `image_url` parts, and we're back at the request DeepSeek rejects (the #432 bug). What I'd do: keep the images on the `conv.Append` message so the transcript still has them, and pass `nil` to `SubmitToAgent` when the model is text-only. Same in `releaseQueuedMessage`.

**A failed queued image wipes what the user is typing.** The `ProcessImageRefs` error branch in `releaseQueuedMessage` calls `m.userInput.Reset()`, and that runs mid-stream — exactly when someone is typing their next message. Just dropping the call fixes it. While you're in there, `buildUserMessage` (the way `drainInputQueueWhileIdle` does it) would also stop the queued item from grabbing images pasted after it was enqueued.

Smaller stuff, take or leave: `adaptImagesForModel` has no test — one would have caught the first point; the cleanup loop in `model_agent_events.go` landed in the middle of an existing comment; and still open from last round, does DeepSeek v4 actually accept `low` and `medium` for `reasoning_effort`?

hsain9357:
Hello @yanmxa  thanks for the review, you can fix the remaining issues in this PR or cherry pick  the necessary commits please.

yanmxa:
Answered the open question: `medium` is not a value the API takes — https://api-docs.deepseek.com/guides/thinking_mode/

Dropped it, kept `low`, added the `xhigh` that was missing (`1371b87d`).

yanmxa:
Please take a look @hchenxa ~

hchenxa (approved):

yanmxa:
Reviewed at `47d0fad9` — build, `go test ./...` and `make lint` are clean, and the eight points from the first round plus the two from the second all check out. The design holds up: all three send paths (`dispatchSubmission`, `releaseQueuedMessage`, and `drainInputQueueWhileIdle`, which routes back through `dispatchSubmission`) go through `adaptTurnForProvider`, `conv` keeps the images while the provider doesn't, and the `/compact` and hint chains are covered.

Three things left. Rather than describe them, I've put them in a PR against this branch so they land inside #439 and keep your authorship: **hsain9357/san#1**.

1. **The inlined temp path outlives nothing.** A clipboard image's temp file is deleted at `OnTurnEnd`, but its path is baked into content that `conv` persists and replays. From turn 2 on, the model reads that path out of its own history and gets `ENOENT`, while the text next to it still says "The files are on disk". Changed to session-scoped cleanup at the quit seam in `run.go`.
2. **`ProcessImageRefs`'s error contract isn't honored by either caller.** The leading-path branch strips the failed path "so the caller can surface a notice while the remaining text survives" — but `buildUserMessage` drops the message, and `releaseQueuedMessage` falls back to `item.Content`, sending the raw leading path. Made the contract true instead of special-casing the caller.
3. **Nits.** `extForMediaType` reverse-walks a map where two extensions map to `image/jpeg`, so the temp suffix flips between `.jpg` and `.jpeg` run to run; plus a dead assignment and a doubled `TrimSpace` in the leading-path branch.

Merge that one into your branch and it flows straight into this PR — or take just the pieces you agree with, no objection either way. Point 1 is the one I'd not ship without.

Everything else I checked and left alone: the `OnTurnEnd` cleanup ordering relative to `drainTurnQueues` is correct (and survives the `StopCancelled` early return), `bareImagePathRe` requires a separator so bare filenames aren't swallowed, `DefaultThinkingEffort` stays `off`, and `thinking: {"type": "disabled"}` matches how `moonshot` and `bigmodel` already send their toggle.

yanmxa:
Heads-up: this merged as `fe74e72b` while my follow-up PR was open against the branch, so I closed that one and reopened the same three fixes against `main` — #457. Point 1 (the temp file being deleted at `OnTurnEnd` while its path stays in the persisted content) is the one worth landing; the other two are a contract cleanup and a nit.

PR #446, opened 2026-08-01
fix: a PreToolUse hook's allow no longer bypasses the permission gate

### LINKED ISSUE #441: A PreToolUse hook's "allow" bypasses deny rules and the circuit breaker
A `PreToolUse` hook that answers `permissionDecision: "allow"` causes the permission gate to be skipped **entirely**, so the call never reaches `HasPermissionToUseTool`. That means a hook can execute a command covered by an explicit user deny rule, and can also get past the circuit breaker.

`internal/setting/permission.go` already documents the intended invariant, and even implements it — but the implementation is never called.

## Where it goes wrong

`internal/tool/pretool_hook.go:99`

```go
if pt.check != nil && (!allowByHook || forceAsk) {
    if allow, reason := pt.check.Check(ctx, pt.inner.Name(), input, forceAsk, permissionReason); !allow {
        return "", fmt.Errorf("blocked: %s", reason)
    }
}
return pt.inner.Execute(ctx, input)
```

When `allowByHook && !forceAsk`, `pt.check.Check` is skipped and the tool runs.

`pt.check` is the `PermissionGate`, and it is the only route to `HasPermissionToUseTool`. The main agent wraps its tools exactly once (`internal/agent/build.go:140`):

```go
Tools: tool.WithPreToolUseAndPermission(tools, p.HookEngine, pg),
```

There is no second `tool.WithPermission` layer behind it, so nothing else re-checks the call.

## The invariant that exists but is not wired

`internal/setting/permission.go:226`

```go
// ResolveHookAllow checks if a hook's "allow" decision should be honored.
// Returns false if a deny rule, a safety check (either tier), or an explicit
// ask rule overrides the hook's decision. This implements the safety
// invariant: deny rules > safety checks > ask rules > hook allow.
```

`ResolveHookAllow` is referenced only by the `Settings` wrapper (`internal/setting/service.go:211`) and by its own unit test (`internal/setting/permission_test.go:1038`). It is never called from the permission path:

```console
$ grep -rn 'ResolveHookAllow' --include='*.go' .
internal/setting/service.go:211:func (s *Settings) ResolveHookAllow(...)
internal/setting/service.go:217:	return s.data.ResolveHookAllow(...)
internal/setting/permission_test.go:1038:func TestResolveHookAllow(t *testing.T) {
internal/setting/permission_test.go:1122:	got := settings.ResolveHookAllow(...)
internal/setting/permission.go:226:// ResolveHookAllow checks if ...
internal/setting/permission.go:230:func (s *Data) ResolveHookAllow(...)
```

The current behaviour is also locked in by `internal/tool/pretool_hook_test.go:76`, which asserts the checker must *not* run after a hook allow — so the two halves of the codebase disagree about the contract.

## What a hook can currently get past

- explicit **deny rules** from settings
- the **circuit breaker** — even though `internal/setting/permission.go:549` calls it "the only check bypass mode cannot skip"
- both **confirmation tiers** (unrecoverable and recoverable)
- explicit **ask rules**

Note the ask-vs-allow conflict *is* handled correctly (`ForceAsk` wins over `PermissionAllow`, `pretool_hook_test.go:117`). It is only the allow path that skips the gate.

## Reproduction

Dropping this in `internal/agent/` fails on both cases today:

```go
type ranTool struct{ ran bool }

func (r *ranTool) Name() string            { return "Bash" }
func (r *ranTool) Description() string     { return "" }
func (r *ranTool) Schema() core.ToolSchema { return core.ToolSchema{Name: "Bash"} }
func (r *ranTool) Execute(context.Context, map[string]any) (string, error) {
	r.ran = true
	return "EXECUTED", nil
}

type allowHook struct{}

func (allowHook) HasHooks(hook.EventType) bool               { return true }
func (allowHook) ExecuteAsync(hook.EventType, hook.HookInput) {}
func (allowHook) StopHookActive() *bool                       { return nil }
func (allowHook) Execute(context.Context, hook.EventType, hook.HookInput) hook.HookOutcome {
	return hook.HookOutcome{ShouldContinue: true, PermissionAllow: true, HookSource: "PreToolUse"}
}

func gateFor(d *setting.Data) *PermissionGate {
	return NewPermissionGate(func(name string, args map[string]any) PermDecisionResult {
		dec := d.HasPermissionToUseTool(name, args, nil)
		return PermDecisionResult{Decision: dec.Behavior, Reason: dec.Reason}
	})
}

func TestHookAllowVsDenyRule(t *testing.T) {
	d := &setting.Data{Permissions: setting.PermissionSettings{Deny: []string{"Bash(rm:*)"}}}
	inner := &ranTool{}
	tools := tool.WithPreToolUseAndPermission(core.NewTools(inner), allowHook{}, gateFor(d))
	tools.Get("Bash").Execute(context.Background(), map[string]any{"command": "rm -rf important/"})
	if inner.ran {
		t.Errorf("hook allow executed a command covered by an explicit deny rule")
	}
}

func TestHookAllowVsCircuitBreaker(t *testing.T) {
	inner := &ranTool{}
	tools := tool.WithPreToolUseAndPermission(core.NewTools(inner), allowHook{}, gateFor(&setting.Data{}))
	tools.Get("Bash").Execute(context.Background(), map[string]any{"command": "rm -rf ~"})
	if inner.ran {
		t.Errorf("hook allow executed a command the circuit breaker should hold")
	}
}
```

Observed:

```
deny rule Bash(rm:*)  -> ran=true out="EXECUTED" err=<nil>
--- FAIL: TestHookAllowVsDenyRule
--- FAIL: TestHookAllowVsCircuitBreaker
```

## Suggested fix

In `preToolHookTool.Execute`, don't let `allowByHook` skip the gate outright — gate it on `ResolveHookAllow` first:

```go
if pt.check != nil && (!allowByHook || forceAsk || !pt.resolveHookAllow(name, input)) {
    ...
}
```

That keeps the hook's usefulness (it can still waive the routine "do you want to run this?" prompt) while leaving deny rules, the breaker, both confirmation tiers and explicit ask rules untouched — exactly what the `ResolveHookAllow` doc comment describes.

`internal/tool/pretool_hook_test.go:76` (`TestPreToolUseAllowOverridesPermissionPrompt`) needs updating alongside, since it currently pins the bypass. Worth adding the two cases above as regression tests: this is easy to reintroduce, because the bypass is the shorter, more obvious way to write the condition and every functional test still passes with it.

yanmxa:
## Problem

A `PreToolUse` hook answering `permissionDecision: "allow"` skipped `pt.check`
entirely, and `pt.check` is the only route to `HasPermissionToUseTool`. A hook
could therefore run a command covered by an explicit deny rule, and get past
the circuit breaker, both confirmation tiers, and explicit ask rules.

`setting.ResolveHookAllow` already documented and implemented the intended
invariant — deny rules > safety checks > ask rules > hook allow — but nothing
ever called it.

Reproduced on `main` before the fix, driving the real hook + gate + settings
stack:

```
deny rule Bash(rm:*)     -> ran=true out="EXECUTED" err=<nil>
circuit breaker rm -rf ~ -> ran=true out="EXECUTED" err=<nil>
```

## Fix

- `PermissionGate` gains `HonorsHookAllow`, backed by a `PermHookAllowFunc` set
  from `BuildParams.HookAllowResolver`; the app supplies
  `Settings.ResolveHookAllow`.
- `preToolHookTool.Execute` consults it, so a hook allow waives only the routine
  prompt — a deny rule, the circuit breaker, either confirmation tier or an
  explicit ask rule still sends the call through the gate.

The waiver fails closed wherever there is nothing to certify it with:

- a nil resolver — a gate with no rules to check the hook against cannot certify
  the waiver, so every call is gated;
- a `Settings` handle with no data loaded — the package default before
  `Initialize` runs, and what `ResetDefaultSettings` restores. It used to return
  `true`, which was the one state where the hook waived the gate with *zero*
  checks behind it, while the same call without a hook would have reached the
  user (`HasPermissionToUseTool` reads nil data as `Prompt`).

Forgetting to wire the resolver, or asking before settings load, costs a
redundant prompt — not a bypass.

## Where the invariant holds

`WithPreToolUseAndPermission` has one call site — the main agent's tool registry — and MCP tools are registered into that same registry before it is wrapped, so they get identical protection. Subagents (`tool.WithPermission`) and the self-learning fork never reach this code: they fire `SubagentStart`/`SubagentStop` but never `PreToolUse` for a tool call, so there is no hook verdict to bypass anything with. Worth stating precisely — on those paths the bug class is absent rather than defended, so wiring `PreToolUse` into them later would need this check wired too.

## Deliberately unchanged

`ResolveHookAllow` does not consider `ModeReadOnly` / `ModeDontAsk`, so a hook
allow still waives those. That matches the invariant the function documents;
widening it is a separate decision, tracked in #452.

`ResolveHookAllow` also keeps its own copies of the deny-rule and ask-rule loops
rather than sharing them with `HasPermissionToUseTool`. That duplication is
pre-existing, but this PR is what puts it on the enforcement path — a tier added
to one and not the other would silently reopen this bug class. Folding the two
into a single pipeline needs a new parameter threaded through `Check`, so it is
tracked separately in #454.

## Tests

- `TestPreToolUseAllowOverridesPermissionPrompt` pinned the bypass. It now covers
  the honored case, with `TestPreToolUseAllowStillGatedWhenNotHonored` for the
  refused one.
- `internal/agent/hook_allow_test.go` covers the seam the bypass lived in: the
  deny rule, the circuit breaker (which now reaches the prompt and is refused),
  and the case that must keep working — a routine call where the hook allow
  still skips the prompt. That last one drives `touch marker`: a read-only
  command like `git status` is permitted by the mode default whether or not the
  waiver works, so it would pin nothing.
- `TestSettingsResolveHookAllowFailsClosedWithoutData` covers the nil-data
  handle, alongside the `HasPermissionToUseTool` result it now agrees with.
- `go build ./...`, `go test ./...` and `make lint` (vet + format + layercheck)
  are clean.

Fixes #441

### DIFF
--- a/internal/agent/hook_allow_test.go
+++ b/internal/agent/hook_allow_test.go
@@ -1,0 +1,127 @@
+package agent
+
+import (
+	"context"
+	"testing"
+
+	"github.com/genai-io/san/internal/core"
+	"github.com/genai-io/san/internal/hook"
+	"github.com/genai-io/san/internal/setting"
+	"github.com/genai-io/san/internal/tool"
+)
+
+// These tests drive the whole hook → gate → settings path the main agent is
+// built with, because the bypass they guard against lived in the seam between
+// those three: a PreToolUse hook answering "allow" used to skip the gate
+// outright, so the call never reached HasPermissionToUseTool.
+
+type recordingTool struct{ ran bool }
+
+func (r *recordingTool) Name() string            { return "Bash" }
+func (r *recordingTool) Description() string     { return "" }
+func (r *recordingTool) Schema() core.ToolSchema { return core.ToolSchema{Name: "Bash"} }
+func (r *recordingTool) Execute(context.Context, map[string]any) (string, error) {
+	r.ran = true
+	return "executed", nil
+}
+
+// allowingHook answers every PreToolUse with "allow", the way a permissive user
+// hook does.
+type allowingHook struct{}
+
+func (allowingHook) HasHooks(hook.EventType) bool                { return true }
+func (allowingHook) ExecuteAsync(hook.EventType, hook.HookInput) {}
+func (allowingHook) StopHookActive() *bool                       { return nil }
+func (allowingHook) Execute(context.Context, hook.EventType, hook.HookInput) hook.HookOutcome {
+	return hook.HookOutcome{ShouldContinue: true, PermissionAllow: true, HookSource: "PreToolUse"}
+}
+
+// gatedTools wires a tool behind the same hook + gate stack buildAgent uses.
+func gatedTools(t *testing.T, data *setting.Data, inner core.Tool) (core.Tools, *PermissionGate) {
+	t.Helper()
+	gate := NewPermissionGate(func(name string, args map[string]any) PermDecisionResult {
+		decision := data.HasPermissionToUseTool(name, args, nil)
+		return PermDecisionResult{Decision: decision.Behavior, Reason: decision.Reason}
+	})
+	gate.SetHookAllowResolver(func(name string, args map[string]any) bool {
+		return data.ResolveHookAllow(name, args, nil)
+	})
+	t.Cleanup(gate.Close)
+	return tool.WithPreToolUseAndPermission(core.NewTools(inner), allowingHook{}, gate), gate
+}
+
+// runBash executes one command through the stack and reports whether the gate
+// prompted on the way. Racing the prompt against the call's own return is what
+// keeps every regression a clean failure: a bypass returns with prompted=false
+// instead of parking the test on a prompt that never arrives, and a prompt is
+// answered (deny) instead of parking the call on an answer that never comes.
+func runBash(t *testing.T, tools core.Tools, gate *PermissionGate, command string) (prompted bool, err error) {
+	t.Helper()
+	done := make(chan error, 1)
+	go func() {
+		_, execErr := tools.Get("Bash").Execute(context.Background(), map[string]any{"command": command})
+		done <- execErr
+	}()
+	prompts := make(chan *PermGateRequest, 1)
+	go func() {
+		if req, ok := gate.Recv(); ok {
+			prompts <- req
+		}
+	}()
+
+	select {
+	case err = <-done:
+		return false, err
+	case req := <-prompts:
+		req.Response <- PermGateResponse{Allow: false, Reason: "denied by test"}
+		return true, <-done
+	}
+}
+
+func TestHookAllowCannotWaiveDenyRule(t *testing.T) {
+	data := &setting.Data{Permissions: setting.PermissionSettings{Deny: []string{"Bash(rm:*)"}}}
+	inner := &recordingTool{}
+	tools, gate := gatedTools(t, data, inner)
+
+	_, err := runBash(t, tools, gate, "rm -rf important/")
+	if err == nil {
+		t.Error("expected the deny rule to block the call despite the hook allow")
+	}
+	if inner.ran {
+		t.Error("hook allow executed a command covered by an explicit deny rule")
+	}
+}
+
+func TestHookAllowCannotWaiveCircuitBreaker(t *testing.T) {
+	inner := &recordingTool{}
+	tools, gate := gatedTools(t, &setting.Data{}, inner)
+
+	prompted, err := runBash(t, tools, gate, "rm -rf ~")
+	if !prompted {
+		t.Error("hook allow skipped the gate: the circuit breaker never reached the user")
+	}
+	if err == nil {
+		t.Error("expected the refused prompt to block the call")
+	}
+	if inner.ran {
+		t.Error("hook allow executed a command the circuit breaker holds")
+	}
+}
+
+// The hook stays useful: on a call no rule and no safety check speaks to, its
+// allow still waives the routine "do you want to run this?" prompt.
+func TestHookAllowWaivesRoutinePrompt(t *testing.T) {
+	inner := &recordingTool{}
+	tools, gate := gatedTools(t, &setting.Data{}, inner)
+
+	prompted, err := runBash(t, tools, gate, "git status")
+	if err != nil {
+		t.Fatalf("Execute returned error: %v", err)
+	}
+	if prompted {
+		t.Error("hook allow should have waived the routine prompt")
+	}
+	if !inner.ran {
+		t.Error("the waived call should have run")
+	}
+}
--- a/internal/agent/hook_allow_test.go
+++ b/internal/agent/hook_allow_test.go
@@ -51,10 +51,11 @@ func gatedTools(t *testing.T, data *setting.Data, inner core.Tool) (core.Tools, 
 }
 
 // runBash executes one command through the stack and reports whether the gate
-// prompted on the way. Racing the prompt against the call's own return is what
-// keeps every regression a clean failure: a bypass returns with prompted=false
-// instead of parking the test on a prompt that never arrives, and a prompt is
-// answered (deny) instead of parking the call on an answer that never comes.
+// prompted on the way. Racing the gate's request channel against the call's own
+// return is what keeps every regression a clean failure: a bypass returns with
+// prompted=false instead of parking the test on a prompt that never arrives,
+// and a prompt is answered (deny) instead of parking the call on an answer that
+// never comes.
 func runBash(t *testing.T, tools core.Tools, gate *PermissionGate, command string) (prompted bool, err error) {
 	t.Helper()
 	done := make(chan error, 1)
@@ -62,17 +63,11 @@ func runBash(t *testing.T, tools core.Tools, gate *PermissionGate, command strin
 		_, execErr := tools.Get("Bash").Execute(context.Background(), map[string]any{"command": command})
 		done <- execErr
 	}()
-	prompts := make(chan *PermGateRequest, 1)
-	go func() {
-		if req, ok := gate.Recv(); ok {
-			prompts <- req
-		}
-	}()
 
 	select {
 	case err = <-done:
 		return false, err
-	case req := <-prompts:
+	case req := <-gate.requests:
 		req.Response <- PermGateResponse{Allow: false, Reason: "denied by test"}
 		return true, <-done
 	}
--- a/internal/agent/hook_allow_test.go
+++ b/internal/agent/hook_allow_test.go
@@ -109,12 +109,14 @@ func TestHookAllowCannotWaiveCircuitBreaker(t *testing.T) {
 }
 
 // The hook stays useful: on a call no rule and no safety check speaks to, its
-// allow still waives the routine "do you want to run this?" prompt.
+// allow still waives the routine "do you want to run this?" prompt. The command
+// has to be one the gate would really prompt on — a read-only one is permitted
+// by the mode default whether or not the waiver works, so it would pin nothing.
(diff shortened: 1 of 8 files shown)


internal/app/agent.go
```diff
@@ -268,6 +268,10 @@ func (m *model) buildAgentParams() agent.BuildParams {
 			return agent.PermReviewResult{Allow: true, Reason: "autopilot: " + verdict.Reason}
 		},
 
+		HookAllowResolver: func(name string, args map[string]any) bool {
```
hchenxa, line 271: This wires `Settings.ResolveHookAllow` in as the enforcement seam, but that function fails **open** when settings aren't loaded: `Settings.ResolveHookAllow` returns `true` when `s.data == nil`, and `&Settings{}` (nil data) is the package default until `Initialize` runs — `ResetDefaultSettings`/`SetDefaultSettings(nil)` restore it. Note its sibling `Settings.HasPermissionToUseTool` treats nil data as `Prompt` ("default: no settings loaded"), so the same command without the hook would at least reach the user — with a hook answering "allow", it skips the gate entirely with zero checks.

That's the exact opposite of the fail-closed principle stated in this PR ("a gate with no rules to check the hook against cannot certify the waiver"). Suggest flipping the nil-data branch to `return false` (one line, plus a unit test on `Settings.ResolveHookAllow` with a nil-data handle).

internal/agent/hook_allow_test.go
```diff
@@ -0,0 +1,127 @@
...
+	}
+}
+
+// The hook stays useful: on a call no rule and no safety check speaks to, its
+// allow still waives the routine "do you want to run this?" prompt.
+func TestHookAllowWaivesRoutinePrompt(t *testing.T) {
+	inner := &recordingTool{}
+	tools, gate := gatedTools(t, &setting.Data{}, inner)
+
+	prompted, err := runBash(t, tools, gate, "git status")
```
hchenxa: `git status` is classified as a read-only Bash command, so the gate's mode default permits it without ever prompting (`ModeDefaultForCall` → "mode: read-only bash command"). `prompted` is therefore false whether or not the waiver works — this test passes with `HonorsHookAllow` removed entirely, and equally if every call were re-routed through the gate. It can't fail on the regression it claims to guard.

Use a command that would actually prompt in the gate but triggers no deny rule, circuit breaker, confirmation tier or ask rule — e.g. `touch marker` (not read-only, ModeDefault → Prompt, no confirmation reason, `ResolveHookAllow` → true) — so the hook allow is the only thing standing between the call and a prompt.

internal/agent/permission.go
```diff
@@ -79,6 +86,25 @@ func (pg *PermissionGate) SetReviewer(fn PermReviewFunc) {
...
+// every call goes through the gate.
+func (pg *PermissionGate) SetHookAllowResolver(fn PermHookAllowFunc) {
+	pg.hookAllowFn = fn
+}
+
+// HonorsHookAllow reports whether a PreToolUse hook's "allow" is enough to skip
+// the gate for this call. A deny rule, the circuit breaker, either confirmation
+// tier or an explicit ask rule outranks the hook, and the call goes through the
+// gate as if the hook had said nothing.
+func (pg *PermissionGate) HonorsHookAllow(name string, input map[string]any) bool {
```
hchenxa, line 110: Question (low priority, flagged as deliberate in the PR description): `ResolveHookAllow` never consults `session.Mode` or `session.ShouldAvoidPrompts`, so in read-only / don't-ask / headless sessions a permissive hook still waives the gate for calls the mode would otherwise reject — e.g. a `Write` inside the working directory (no sensitive path → no confirmation reason) runs in read-only mode, where the gate's mode default would have said "mode: read-only". This is pre-existing (a hook allow used to skip the gate in every mode), so it's not a regression — but the PR frames the invariant as "a hook cannot vouch for a call the user or the breaker has already ruled on", and the mode's ruling looks like the same class. Consider a follow-up: should mode veto the waiver?

hchenxa (commented):
Solid fix — the seam was real, and the "deny rules > safety checks > ask rules > hook allow" invariant now actually holds on the main agent path. The fail-closed nil-resolver design and the new tests around the deny rule / circuit breaker are good. Three comments:

1. The resolver the app wires can fail **open** when settings are nil, contradicting the PR's own fail-closed principle.
2. The happy-path test uses a command the gate never prompts on, so it pins nothing.
3. A question about read-only / don't-ask sessions, which remain waivable by a hook allow (flagged as deliberate — asking whether that should follow up).

internal/app/agent.go
yanmxa: Fixed in 0f69603b — `Settings.ResolveHookAllow` now returns `false` on a nil data handle.

You're right that this was the one state where the waiver was certified by nothing at all, and the contrast with `HasPermissionToUseTool` reading the same state as `Prompt` makes it worse: the hook turned a call the user would have been asked about into one that ran unasked. Flipped it, with the reasoning in a comment at the branch, and added `TestSettingsResolveHookAllowFailsClosedWithoutData` — it pins both halves, the refused waiver and the `Prompt` the same call gets without the hook.

internal/agent/hook_allow_test.go
yanmxa: Fixed in 0f69603b — the test drives `touch marker` now.

Confirmed the test was vacuous as written: with `git status` it passes with the waiver removed. I checked by replacing `pt.check.HonorsHookAllow(...)` with `false` in `preToolHookTool.Execute` and rerunning —

- `git status`: still passes (the mode default permits it, no prompt either way);
- `touch marker`: fails with `Execute returned error: blocked: denied by test`, i.e. the call reached the gate and the test's refusal blocked it.

So the hook allow is now the only thing between that call and a prompt, which is what the test claims to guard. Added a note above it so the next person doesn't reach for a read-only command.

internal/agent/permission.go
yanmxa: Good question, and I agree the mode's ruling is the same class — the user said what this session may do before any call existed, which is exactly the kind of prior decision the invariant says a hook cannot vouch for.

Keeping it out of this PR: it is not a regression (a hook allow skipped the gate in every mode before), and the answer isn't uniform across modes. Read-only is the clear case. `ModeDontAsk` / headless is murkier — the gate answers `Reject` there rather than prompting, so vetoing the waiver turns "the hook allowed it" into a hard failure instead of a prompt, which may not be what someone writing an allow hook for an unattended run expects.

Filed #452 with that trade-off and a sketch of the change; linked it from the "Deliberately unchanged" section so the PR no longer just asserts the scope.

yanmxa:
@hchenxa gentle ping — this is ready for another look.

Your three comments are addressed in 0f69603b:

1. **Fail-open on nil settings** — flipped `Settings.ResolveHookAllow` to `false` for a nil data handle, with `TestSettingsResolveHookAllowFailsClosedWithoutData` covering both halves (the refused waiver, and the `Prompt` the same call gets without the hook).
2. **Vacuous happy-path test** — confirmed it: with `git status` the test passes even with `HonorsHookAllow` removed. Switched to `touch marker`, which fails with `blocked: denied by test` when the waiver is stubbed out, so the hook allow is now the only thing between that call and a prompt.
3. **Mode vs. hook allow** — agreed the mode's ruling is the same class, but the answer isn't uniform across modes (read-only is clear; `ModeDontAsk`/headless would turn "the hook allowed it" into a hard failure rather than a prompt). Filed #452 with the trade-off and a sketch, linked from the description.

CI is green on both `test` and `test-windows`.

PR #449, opened 2026-08-02
fix: report a call waiting on a permission prompt as waiting, not running

### LINKED ISSUE #440: Permission-gated tool calls are stamped as started before they are allowed to run
## Problem

A tool call is stamped as started by its `PreToolEvent` (`internal/app/conv/update.go`, `applyPreTool` → `MarkCurrent` + `MarkStarted`), which fires *before* the permission request goes out. The call has not been allowed to run yet, so:

- its row spins with a live elapsed timer while the user is deciding — the timer counts the user's deliberation, not execution;
- nothing in the render path can tell that call apart from a batch sibling that really is executing (the permission gate lives inside each tool's `Execute` — `internal/tool/permission.go:51` — and a parallel batch runs one goroutine per call, `internal/core/agent_impl.go:535-542`).

#437 works around this by freezing *every* in-flight row while a docked modal is up (`RenderContext.DockedModalActive`). That is deliberately blunt: in a mixed batch `[A allow-listed, B prompts]`, A genuinely runs while B's modal is up, but A's row shows a static bullet with no timer until B is answered.

## Proposed fix

Give the pending call an identity on the permission request so the view can name exactly which row is waiting:

- add `ToolCallID` to `PermGateRequest` (`internal/agent/permission.go`) — `prompt` already runs under the ctx `core.WithToolCallID` stamped in `internal/core/agent_impl.go`;
- record it as `ToolExecState.AwaitingApproval` in `OnPermGateRequest` (`internal/app/agent.go`);
- clear it and `MarkStarted` on approval in `handlePermGateDecision` (`internal/app/update_approval.go`), so the elapsed timer measures execution rather than deliberation;
- have `toolCallIcon` / `runningRowDetail` consult that state instead of the blanket `DockedModalActive` freeze.

Siblings that are genuinely running keep spinning and ticking; the row being approved gets an honest "waiting on you" state rather than borrowing the queued-row bullet.

Also worth folding in: `renderAndCommit` (`internal/app/model_scrollback.go`) builds its params without the modal flag, so a resize reflow or compact reprint while a modal is up can freeze a spinning row into immutable scrollback. Pre-existing, but the same state would fix it.

yanmxa:
Fixes #440.

## Problem

`applyPreTool` stamps a tool call as started (`MarkCurrent` + `MarkStarted`) on its `PreToolEvent`, which the agent emits in the *resolve* phase — before the tool's `Execute` runs, and therefore before the permission request goes out. A gated call was already "running" as far as the view was concerned:

- its row spun with a live elapsed timer while the user was still deciding, so the timer measured deliberation, not execution;
- nothing in the render path could tell that call apart from a batch sibling that really was executing — the permission gate lives inside each tool's `Execute`, and a parallel batch runs one goroutine per call.

#437 worked around it by freezing *every* in-flight row while a docked modal is up (`RenderContext.DockedModalActive`). In a mixed batch `[A allow-listed, B prompts]`, A genuinely runs while B's modal is open, but A's row showed a static bullet with no timer until B was answered.

## Fix

Give the pending call an identity on the permission request, so the view can name exactly which row is waiting:

- `PermGateRequest` carries `ToolCallID`, stamped in `prompt()` from the context the agent puts on every call (`core.WithToolCallID`) — the permission func signature carries the tool name, not the call. Both entry points get it: the normal decider path and a hook-forced prompt.
- `HandlePermGate` records it as `ToolExecState.AwaitingApprovalID`; `handlePermGateDecision` clears it, and on approval calls `MarkStarted` again so the row's timer measures execution instead of the wait for an answer.
- `toolCallIcon` / `runningRowDetail` and the Agent row's blink consult that state. The gated row holds a static bullet and says `· waiting for approval`; siblings that were allowed to run keep spinning and ticking.
- The state is per batch: `Track` and `Reset` clear it, so a new batch or a cancelled turn never inherits a stale waiting row.

Naming a call is precise but not universal, so the blanket freeze stays as the fallback rather than being dropped (`modalNamesNoCall`): when a docked modal is up and **no** call is named, every in-flight row holds still exactly as it did before this PR — static bullet, no row detail, no blinking Agent row, and no "waiting for approval" claim nobody made. That covers the modals that park a call without having an ID to hand over — an `AskUserQuestion` prompt is the live case, a secret prompt the other — and doubles as the safety net for a permission request that somehow arrives with an empty `ToolCallID` (`HandlePermGate` logs that case, so the degradation is visible rather than silent).

`DockedModalActive` keeps the two freezes that really are about the modal: the assistant bullet (the turn is parked on the answer) and the `(ctrl+o to expand)` hint (the overlay owns the keyboard).

The scrollback concern in the issue comes along for free: every render path — live view, `renderAndCommit`, background flush — builds its params through `messageRenderParams`, so a resize reflow or compact reprint while a prompt is open now commits the waiting state rather than a spinner frame for a call that was never allowed to start.

## Not covered

The fallback is a stopgap, not the design: only the permission modal names its call, so a genuinely running sibling still freezes while an `AskUserQuestion` is open. The deeper fix — thread the call ID through every docked modal, at which point the fallback stops firing on its own — is #453; the ID is already in the context at each of those seams, it just isn't forwarded.

A call *queued* behind an open prompt is still drawn as running. Two shapes reach it: a second gated call in a parallel batch, whose request is sitting in the gate channel the TUI has not polled yet; and later calls of a sequential batch, which the view already renders as in flight whenever a batch has more than one call (`ParallelMode: len(PendingCalls) > 1`). Both predate this change and need the agent to tell the UI which calls actually began executing — worth a separate issue.

## Tests

- `internal/agent`: the gate stamps `ToolCallID` from the execution context, on the decider path and on a hook-forced prompt.
- `internal/app/conv`: a gated row reports waiting while its running sibling keeps its spinner and timer; an Agent row awaiting approval stops blinking; a docked modal that names no call freezes every row without claiming an approval was asked for; approving restamps the call and clears the waiting state; `Track`/`Reset` clear it.
- `internal/app`: the modal frame no longer shows the gated call as running, `HandlePermGate` marks the call its request names, and a request with no ID leaves the per-call state empty so the rows fall back to the modal freeze. The `ctrl+o` freeze keeps its own test on the modal flag.

Manual check: run a `Bash` call that prompts, leave the modal up for ~20s, approve — the row starts its timer at approval instead of showing 20s+.

### DIFF
--- a/internal/app/agent.go
+++ b/internal/app/agent.go
@@ -573,8 +573,13 @@ func (m *model) ContinueOutbox() tea.Cmd {
 func (m *model) HandlePermGate(req *conv.PermGateRequest) tea.Cmd {
 	m.services.Agent.SetPendingPermission(req)
 	if req == nil {
+		m.conv.Tool.ClearAwaitingApproval()
 		return nil
 	}
+	// The call was stamped as started by its PreToolEvent, which fired before
+	// this request: name it so its row reports waiting on the user while its
+	// batch siblings keep reporting the work they really are doing.
+	m.conv.Tool.MarkAwaitingApproval(req.ToolCallID)
 
 	permReq := m.preparePermissionRequest(req)
 	// Emit permission.required with the metadata about to be rendered to the
--- a/internal/app/agent.go
+++ b/internal/app/agent.go
@@ -578,7 +578,15 @@ func (m *model) HandlePermGate(req *conv.PermGateRequest) tea.Cmd {
 	}
 	// The call was stamped as started by its PreToolEvent, which fired before
 	// this request: name it so its row reports waiting on the user while its
-	// batch siblings keep reporting the work they really are doing.
+	// batch siblings keep reporting the work they really are doing. A request
+	// with no ID names nothing, which leaves the rows on the blunt freeze the
+	// modal already forces (see conv.modalNamesNoCall) — correct, but only
+	// because it degrades, so say so rather than let a future caller lose the
+	// per-call state without a trace.
+	if req.ToolCallID == "" {
+		log.Logger().Warn("permission request carries no tool call ID; every in-flight row falls back to the docked-modal freeze",
+			zap.String("tool", req.ToolName))
+	}
 	m.conv.Tool.MarkAwaitingApproval(req.ToolCallID)
 
 	permReq := m.preparePermissionRequest(req)
--- a/internal/app/conv/message.go
+++ b/internal/app/conv/message.go
@@ -502,16 +502,19 @@ type ToolCallsParams struct {
 	// ToolProgress maps a running command's ID to its latest output counter,
 	// shown next to the timer. Empty for calls that have produced no output.
 	ToolProgress map[string]string
-	ModelName    string
-	InputTokens  int
-	OutputTokens int
-	Blink        int
-	AgentColors  map[string]string
-	SpinnerView  string
-	TaskOwnerMap map[string]string
-	MDRenderer   *MDRenderer
-	Width        int
-	Interactive  bool
+	// AwaitingApprovalID is the call parked on a permission prompt: started by
+	// its PreToolEvent, but not allowed to run yet.
+	AwaitingApprovalID string
+	ModelName          string
+	InputTokens        int
+	OutputTokens       int
+	Blink              int
+	AgentColors        map[string]string
+	SpinnerView        string
+	TaskOwnerMap       map[string]string
+	MDRenderer         *MDRenderer
+	Width              int
+	Interactive        bool
 }
 
 // expandHints reports whether "(ctrl+o to expand)" affordances belong on this
@@ -521,6 +524,12 @@ type ToolCallsParams struct {
 // offering the hint.
 func (p ToolCallsParams) expandHints() bool {
 	return p.Interactive && !p.DockedModalActive
+}
+
+// awaitsApproval reports whether this call is the one a permission prompt is
+// asking about — started on paper, but not allowed to run yet.
+func (p ToolCallsParams) awaitsApproval(toolCallID string) bool {
+	return p.AwaitingApprovalID != "" && p.AwaitingApprovalID == toolCallID
 }
 
 // ToolResultData holds the data needed to render a tool result inline.
@@ -565,11 +574,11 @@ func RenderToolCalls(params ToolCallsParams) string {
 			if hasResult {
 				sb.WriteString(renderAgentToolLine(label, params.Width, "●", color) + "\n")
 			} else {
-				// agentIcon blinks ●/○ off the frame counter; under a docked
-				// modal it would advertise a subagent as working while the
-				// turn waits on the user, same as a spinning tool row.
+				// agentIcon blinks ●/○ off the frame counter; on the call a
+				// permission prompt is asking about that would advertise a
+				// subagent as working before it was allowed to spawn.
 				icon := agentIcon(params.Blink)
-				if params.DockedModalActive {
+				if params.awaitsApproval(tc.ID) {
 					icon = "●"
 				}
 				sb.WriteString(renderAgentToolLine(label, params.Width, icon, color))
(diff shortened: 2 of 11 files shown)


internal/app/conv/message.go
```diff
@@ -642,21 +651,17 @@ func RenderToolCalls(params ToolCallsParams) string {
...
+// it holds a static bullet until it is allowed to run. Its batch siblings keep
+// spinning: the gate lives inside each tool's Execute and a parallel batch runs
+// one goroutine per call, so an allow-listed sibling really is executing while
+// the user decides.
 func toolCallIcon(tc core.ToolCall, params ToolCallsParams) string {
 	if _, done := params.ResultMap[tc.ID]; done {
 		return "●"
 	}
-	if params.DockedModalActive {
+	if params.awaitsApproval(tc.ID) {
```
hchenxa: `toolCallIcon`/`runningRowDetail` no longer consult `DockedModalActive`, and the awaiting mechanism only names a call when a *permission* request is open. Under a Question modal (`AskUserQuestion` — `conv.QuestionPrompt` is a docked modal), the asked call is equally parked on the user: its `PreToolEvent` stamped it before the question went out, but no permission request exists to mark it awaiting, so its row now spins and ticks the user's answer time — the exact #440 symptom this PR fixes, via the question path. The old blanket freeze covered this; dropping it is a behavior change the description doesn't mention (the "DockedModalActive keeps the two freezes…" note only covers the bullet and the ctrl+o hint).

Suggestion: keep the conservative freeze when a docked modal is up but no call is named — `params.DockedModalActive && params.AwaitingApprovalID == ""` in `toolCallIcon`/`runningRowDetail`. That restores the old behavior for Question/secret modals and doubles as the fallback for a permission request with an empty `ToolCallID` (see comment on `HandlePermGate`).

internal/app/agent.go
```diff
@@ -573,8 +573,13 @@ func (m *model) ContinueOutbox() tea.Cmd {
 func (m *model) HandlePermGate(req *conv.PermGateRequest) tea.Cmd {
 	m.services.Agent.SetPendingPermission(req)
 	if req == nil {
+		m.conv.Tool.ClearAwaitingApproval()
 		return nil
 	}
+	// The call was stamped as started by its PreToolEvent, which fired before
+	// this request: name it so its row reports waiting on the user while its
+	// batch siblings keep reporting the work they really are doing.
+	m.conv.Tool.MarkAwaitingApproval(req.ToolCallID)
```
hchenxa, line 590: Robustness note: `MarkAwaitingApproval(req.ToolCallID)` with an empty ID behaves exactly like the clear — no row reports waiting, and the gated call's row spins with a timer over the user's deliberation, silently re-introducing #440. Unreachable in the main agent today (`executeToolTask` stamps every tool ctx via `core.WithToolCallID`, and subagents don't use the gate), but this seam fails *silently* rather than conservatively, which is the opposite of the rest of the PR's design. A guard here (`if req.ToolCallID == ""` → log and skip, or fall back to the `DockedModalActive && AwaitingApprovalID == ""` freeze suggested on `toolCallIcon`) would make any future degradation visible instead of quiet.

hchenxa (commented):
Well-scoped fix: the permission request now names the call it gates (`ToolCallID` stamped from the execution context in `prompt()`), the TUI records it as `AwaitingApprovalID`, and the render layer freezes exactly that row — "· waiting for approval" — while batch siblings keep spinning and ticking; approval restamps the timer so it measures execution, and `Track`/`Reset` clear the state per batch. The kept `DockedModalActive` freezes (assistant bullet, ctrl+o hint) are the right call. Two comments:

1. The blanket freeze was dropped from the tool rows, but the replacement only names a call when a *permission* request is open — under a Question modal the asked call regresses to the exact #440 symptom.
2. A request with an empty `ToolCallID` fails silently at the seam that sets the state.

internal/app/conv/message.go
yanmxa: Fixed in 582d605f — took the suggestion as written, behind a named predicate:

```go
func (p ToolCallsParams) modalNamesNoCall() bool {
	return p.DockedModalActive && p.AwaitingApprovalID == ""
}
```

`toolCallIcon` and the Agent row's blink freeze on `awaitsApproval(tc.ID) || modalNamesNoCall()`. `runningRowDetail` keeps the two apart: a named call says `· waiting for approval`, an unnamed one drops the detail — under a Question modal nobody asked for an approval, so claiming one would be its own wrong label. That is also exactly what the old code did (`""`), so the fallback is the pre-PR behavior, not a new one.

You were right that it was a live regression and not just a theoretical one. `TestRenderToolCallsFreezesEveryRowUnderModalNamingNoCall` covers it with an `AskUserQuestion` call and an Agent sibling; with `modalNamesNoCall` stubbed to `false` it fails on `"⋯ AskUserQuestion(0 questions) · 9s"` — the #440 symptom, via the question path, exactly as you described.

Also updated the PR description: the "DockedModalActive keeps…" paragraph now says the blanket freeze survives as the fallback rather than implying the tool rows dropped it.

internal/app/agent.go
yanmxa: Fixed in 582d605f, both halves of your suggestion.

The `modalNamesNoCall` fallback on the render side makes the empty-ID case *conservative* instead of silently wrong: with nothing named, `AwaitingApprovalID` stays empty and every row falls back to the docked-modal freeze rather than spinning over the user's deliberation. And `HandlePermGate` now logs it, so the degradation is visible:

```go
if req.ToolCallID == "" {
	log.Logger().Warn("permission request carries no tool call ID; every in-flight row falls back to the docked-modal freeze",
		zap.String("tool", req.ToolName))
}
```

Kept `MarkAwaitingApproval(req.ToolCallID)` unconditional after the log — an empty ID is exactly the clear, which is the state the fallback wants.

`TestHandlePermGateWithoutCallIDFallsBackToTheModalFreeze` pins it end to end: an ID-less request leaves the state empty and the rendered frame has no spinning tool row. It fails without the fallback (`tool row "✦ Bash(df -h) - check disk" spins while the modal owns the screen`).

yanmxa:
@hchenxa gentle ping — this is ready for another look.

Both of your comments are addressed in 582d605f:

1. **Question modal regression** — took your suggestion as written, behind `modalNamesNoCall()` (`DockedModalActive && AwaitingApprovalID == ""`). You were right that it was live, not theoretical: `TestRenderToolCallsFreezesEveryRowUnderModalNamingNoCall` fails on `⋯ AskUserQuestion(0 questions) · 9s` with the fallback stubbed out. `runningRowDetail` keeps the two cases apart — a named call says `· waiting for approval`, an unnamed one just drops the detail, since nobody asked for an approval under a Question modal.
2. **Empty `ToolCallID`** — the fallback makes it conservative instead of silently wrong, and `HandlePermGate` now logs it so the degradation is visible. `TestHandlePermGateWithoutCallIDFallsBackToTheModalFreeze` pins it end to end.

Also fixed the DCO sign-off (`dco:no` → `dco:yes`) and updated the description, which previously implied the tool rows had dropped the blanket freeze outright.

CI is green on both `test` and `test-windows`.

PR #459, opened 2026-08-04
fix(tui): let shift+tab cycle modes mid-turn and stop `-r` demoting the mode

### LINKED ISSUE #448: shift + tab does not cycle between modes when  run  san -r
### Duplicate check

- [x] I searched existing issues and didn't find a duplicate
- [x] This is a functional bug, not a security vulnerability (those must go through [SECURITY.md](https://github.com/genai-io/san/blob/main/SECURITY.md))

### What happened?

shift + tab does not cycle between auto pilot, bypass permissions etc when switching to old session via `san -r` and pick a session it keep asking me to allow every command which is very annoying. 

### Steps to reproduce

run `san -r`
pick a session 
click on the keyboard shift + tab 

### Environment

- San version:
- OS / arch:
- Go version (if building from source):
- Provider / model:

### Logs

```shell

```

yanmxa:
## What & why

Closes #448

Two defects, stacked:

**`san -r` demotes the permission mode.** `restoreSessionData` treated a session that recorded *no* mode (every session saved before modes were persisted) as one that recorded `"normal"`, overwriting the mode the launch restored from `lastOperationMode`. `san` kept `⏵⏵ accept edits on`; `san -r` + pick dropped to ask-for-everything. Now only a recorded mode moves the user.

**Shift+Tab is dead while a turn runs.** `Stream.Active` stays true for the whole turn (`applyChunk` clears it only on a text-only Done), so the key did nothing during tool calls and approval prompts — the one moment you reach for accept-edits or autopilot — while the status bar still said `(shift+tab to cycle)`. The permission gate reads the posture per call, so a mid-turn switch lands on the next tool. The approval prompt keeps Shift+Tab as its "allow all this session" accelerator (it's an overlay, `routeKeypress` gets it first); the other overlay checks in the guard were dead for the same reason and are gone.

Mid-turn cycling means the UI goroutine rewrites the posture while the agent goroutine reads it, so `autopilotEngaged` / `env.SessionMode` now read the new lock-held `SessionPermissions.CurrentMode()`, and `ResolveHookAllow` snapshots like `HasPermissionToUseTool` already did.

## Type of change

- [x] Bug fix

## Testing

```bash
make lint && go test ./...
CGO_ENABLED=1 go test -race ./internal/app/... ./internal/setting/... ./tests/integration/...
```

Four new tests cover the resume mode, mid-stream cycling, the approval accelerator, and the goroutine race (it reports DATA RACE against the pre-fix reader). Also verified in the real TUI.

## Checklist

- [x] Follows [Conventional Commits](https://www.conventionalcommits.org/)
- [x] All commits signed off (`git commit -s`)
- [x] Tests pass locally
- [x] Docs updated (if behavior or config changed) — none needed; docs already describe Shift+Tab as switching modes mid-session

### DIFF
--- a/internal/app/autopilot.go
+++ b/internal/app/autopilot.go
@@ -107,8 +107,11 @@ func (m *model) liveAutopilotConfig() setting.AutoPilotSettings {
 // the precondition every steer shares. Steers are the copilot's actions, so
 // none fire unless the copilot is actually driving. Combine with the per-steer
 // toggle at each gate: `if !m.autopilotEngaged() || !<steer> { return }`.
+//
+// Read from the synchronized posture, not env.OperationMode: the steer gates
+// run on the agent goroutine while Shift+Tab may be cycling modes mid-turn.
 func (m *model) autopilotEngaged() bool {
-	return m.env.OperationMode == setting.ModeAutoPilot
+	return m.env.SessionPermissions.CurrentMode() == setting.ModeAutoPilot
 }
 
 var (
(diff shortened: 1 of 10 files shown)


hchenxa:
re-trigger

PR #462, opened 2026-08-05
feat: remove the stale workflows and use the san-ci to manage the issue and pr

hchenxa:
the san-ci can do the issue/pr manage in https://github.com/genai-io/san-ci/pull/36, so we can remove this workflow and let san-ci to help on that.


yanmxa:
LGTM

PR #470, opened 2026-08-11
feat(tui): rename bypass permissions to YOLO mode, gate it from /config

yanmxa:
## Why

The unrestricted mode was named from the implementation's point of view — "bypass permissions" — and painted with the error color. The name describes what the code skips rather than what the user chose, and red is the status line's word for "something went wrong", not for a mode you deliberately turned on.

## What changed

- **`⏵⏵ bypass permissions on` → `⏵⏵ YOLO`**, in a new violet theme token so red keeps meaning failure.
- **The three mode labels now read as a set.** Renaming one exposed the drift; `on` is gone from all of them. The rest of the status line is already bare noun phrases (`Opus 5 (high)`, `ctx 12k/200k`), the `⏵⏵` glyph already says the mode is live, and the indicator only renders outside default mode — so `on` carried no information.
- **Thinking effort had two shapes for one fact:** OpenAI models appended it to the model name, every other provider got a separate `✦ high` indicator. Everyone now gets the model-name form; the standalone indicator is gone. `Ctrl+T` stays in the README shortcut table.
- **New `/config › permissions` panel** — the Permissions panel the popup shell already anticipated. `allowBypass` is opt-out, so taking YOLO away previously meant hand-editing `settings.json`.
- **Three fixes so the lock means something:** demotion now goes through `persistOperationMode` (otherwise `lastOperationMode` stayed `"bypass"` and the next launch restored YOLO unprompted); a failed settings reload no longer reports success while the in-memory gate stays stale; a project-level `allowBypass` outranking the user-level write is reported honestly instead of as a successful lock.

## Scope of the gate — stated, not fixed

`allowBypass` is read in exactly two places: `OperationMode.NextWithBypass` and `env.ApplyDefaultPermissionMode`. The subagent path never consults it, so an agent declaring `mode: bypassPermissions` runs unrestricted even with the gate locked. This PR states that in the panel body, the file comment and `permission-model.md` rather than changing it — aligning the behavior changes what a subagent's declared mode means, and belongs in its own PR.

## Compatibility

`allowBypass` and `bypassPermissions` are unchanged on disk and on the wire; only the display name moves. `"yolo"` is accepted as an input alias in both parsers.

`go build ./...`, `go test ./...` and `make lint` pass.

### DIFF
--- a/internal/app/input/config_permissions.go
+++ b/internal/app/input/config_permissions.go
@@ -8,13 +8,11 @@
 // how you take it away without hand-editing settings.json. It persists to
 // the user settings file, matching the appearance panel's user-level scope.
 //
-// Scope, precisely: `allowBypass` is read in exactly two places — the
-// shift+tab cycle (OperationMode.NextWithBypass) and the startup default
-// (env.ApplyDefaultPermissionMode). A subagent that declares
-// `mode: bypassPermissions` resolves through subagent.NormalizePermissionMode,
-// which never consults it, and still runs unrestricted. So this is a lock on
-// how *you* reach the mode, not a session-wide guarantee that nothing runs
-// unprompted — don't describe it as one.
+// Scope: `allowBypass` is read by OperationMode.NextWithBypass and
+// env.ApplyDefaultPermissionMode, and nowhere else — subagent.
+// NormalizePermissionMode never consults it, so an agent declaring
+// `mode: bypassPermissions` runs unrestricted with the gate locked. This
+// gates how you reach the mode, not what the session as a whole may do.
 package input
 
 import (
@@ -40,18 +38,17 @@ type yoloOption struct {
 	allowed bool
 }
 
-func yoloOptions() []yoloOption {
-	return []yoloOption{
-		{label: "Allowed", desc: "shift+tab reaches YOLO mode — no permission prompts", allowed: true},
-		{label: "Locked", desc: "shift+tab stops at autopilot", allowed: false},
-	}
+// yoloOptions is the row list, in display order. Allowed comes first so the
+// cursor parks on the default at Enter.
+var yoloOptions = []yoloOption{
+	{label: "Allowed", desc: "shift+tab reaches YOLO mode — no permission prompts", allowed: true},
+	{label: "Locked", desc: "shift+tab stops at autopilot", allowed: false},
 }
 
 type permissionsPanel struct {
 	settings *setting.Settings
 
-	options []yoloOption
-	cursor  int
+	cursor int
 
 	// baseline is the effective value on disk, marked "● current".
 	baseline bool
@@ -68,21 +65,21 @@ func newPermissionsPanel(settings *setting.Settings) *permissionsPanel {
 func (p *permissionsPanel) Title() string { return "permissions" }
 
 func (p *permissionsPanel) Enter() {
-	p.options = yoloOptions()
 	// Unset means allowed — mirror Settings.AllowBypass's opt-out default so a
 	// fresh install shows "Allowed ● current" instead of a wrong "Locked".
 	p.baseline = p.settings == nil || p.settings.AllowBypass()
-	p.cursor = indexOfYolo(p.baseline)
+	p.cursor = 0
+	for i, opt := range yoloOptions {
+		if opt.allowed == p.baseline {
+			p.cursor = i
+			break
+		}
+	}
 	p.saveErr = nil
 }
 
 func (p *permissionsPanel) Dirty() bool {
-	// Enter() populates options; guard so a Dirty() consulted before it (e.g.
-	// a shell that polls every tab for an unsaved marker) cannot panic.
-	if len(p.options) == 0 {
-		return false
-	}
-	return p.options[p.cursor].allowed != p.baseline
+	return yoloOptions[p.cursor].allowed != p.baseline
 }
 
 func (p *permissionsPanel) HandleKey(msg tea.KeyMsg) (tea.Cmd, bool) {
@@ -93,12 +90,12 @@ func (p *permissionsPanel) HandleKey(msg tea.KeyMsg) (tea.Cmd, bool) {
 		}
 		p.saveErr = nil
 	case "down", "j":
-		if p.cursor < len(p.options)-1 {
+		if p.cursor < len(yoloOptions)-1 {
 			p.cursor++
 		}
 		p.saveErr = nil
 	case "enter", " ":
-		return p.apply(p.options[p.cursor])
+		return p.apply(yoloOptions[p.cursor])
 	}
 	return nil, false
 }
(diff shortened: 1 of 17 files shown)


hchenxa (commented):
Review (medium effort): 5 findings. Two correctness issues in the new `AllowBypassSavedMsg` handler — both stem from acting on the panel's user-level write intent instead of the effective merged value after reload — plus a status-line regression for non-OpenAI providers, one duplication, and one stale comment. Details inline.

internal/app/update.go
```diff
@@ -260,6 +261,36 @@ func (m *model) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
...
+			log.Logger().Warn("reload settings after YOLO-mode save failed", zap.Error(err))
+			m.conv.AddNotice("Saved, but couldn't re-read settings — restart for it to take effect")
+			return m, nil
+		}
+		// Locking the gate has to drop a session that is already in YOLO mode.
+		// Otherwise the switch reads "locked" while the running session keeps
+		// skipping prompts until the next restart. Go through the shared tail
+		// so the new mode is persisted too — otherwise lastOperationMode stays
+		// "bypass" and the next launch restores YOLO behind the user's back.
+		if !msg.Allowed && m.env.OperationMode == setting.ModeBypassPermissions {
```
hchenxa: The demote fires from the panel's write intent (`msg.Allowed`) alone — the effective gate is only consulted two lines down, in the notice switch. If a project-level `allowBypass: true` (in `.san/settings.json`, `.claude/settings.json`, or the `.local` variants) outranks the user-level write — loader priority is project > user, and `mergeSettings`/`coalesceBool` is overlay-wins — the lock did not stick, yet the running session is still dropped to normal and `persistOperationMode()` rewrites user-level `lastOperationMode` from `bypass` to `normal`, while the notice below says "Locked for your user, but project settings still allow YOLO mode".

The user loses both the live mode and the persisted startup default for a lock that never took effect. And because `StartupMode()` is `coalesce(LastOperationMode, Permissions.DefaultMode)` (settings.go:380), the clobbered user-level `lastOperationMode: normal` permanently shadows a project-level `defaultMode: bypassPermissions` on every subsequent launch — even after the project setting is removed.

Suggestion: gate the demote on the effective value — e.g. move it into the `default:` ("YOLO mode locked") branch, after verifying `!m.services.Setting.AllowBypass()` — so an ineffective lock neither demotes the session nor destroys the persisted mode.

internal/app/update.go
```diff
@@ -260,6 +261,36 @@ func (m *model) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
...
+			m.env.OperationMode = setting.ModeNormal
+			m.applyOperationMode()
+			m.persistOperationMode()
+		}
+		switch {
+		case msg.Allowed:
+			m.conv.AddNotice("YOLO mode allowed — shift+tab to reach it")
```
hchenxa, line 290: The lock branches verify the effective post-reload value (`m.services.Setting.AllowBypass()` in the next case), but this allow branch takes the panel's user-level write at face value. If a project-level `allowBypass: false` outranks the user-level `true` this write produces, the notice claims "YOLO mode allowed — shift+tab to reach it" while `cycleOperationMode` → `NextWithBypass(AllowBypass())` still excludes YOLO from the cycle — shift+tab stops at autopilot. Mirror image of the lock branch: check `m.services.Setting.AllowBypass()` here too and report when the user-level allow is outranked (e.g. "Allowed for your user, but project settings still lock it").

internal/app/view.go
```diff
@@ -327,13 +326,7 @@ func (m *model) renderTrackerList() string {
 }
```
hchenxa, line 327: Behavior regression for non-OpenAI providers: thinking effort now rides inside the model segment, which the hook override below (`modelName = status`) replaces for the whole duration a `statusMessage`-bearing hook runs (e.g. a slow `UserPromptSubmit`/`Notification` command hook — `CurrentStatusMessage` is tracked per active hook in internal/hook). Before this PR the effort rendered as an independent left-cluster segment (`✦ high`, never overwritten); now the status line shows neither the effort nor the (also removed) `(ctrl+t to cycle)` affordance while the hook is active.

If the loss is intentional that's fine, but the old OpenAI path was the only one that could already lose the effort this way — the PR extends that to every provider. Consider appending the effort to the status message instead, or leaving it.

internal/app/input/config_permissions.go
```diff
@@ -0,0 +1,179 @@
...
+
+	if p.saveErr != nil {
+		b.WriteString("\n")
+		b.WriteString(appearanceErrorStyle.Render("⚠ couldn't save: " + p.saveErr.Error()))
+		b.WriteString("\n")
+	}
+	return b.String()
+}
+
+func (p *permissionsPanel) renderOption(i int, opt yoloOption) string {
```
hchenxa: `renderOption` (152-168) and `HandleKey` (88-107) are near-verbatim copies of `appearancePanel`'s (`config_appearance.go:203-220` and `117-136`): same ▸/○/● radio rendering, same "current" marker, same `max(8-len(label),1)` label padding, same up/k down/j enter/space switch with `saveErr` cleared on navigation, same `⚠ couldn't save` error render — ~34 duplicated lines differing only in the current-check and the apply/persist target.

Since this panel is deliberately modeled on its sibling, worth extracting a shared radio-row / radio-list helper in `package input` now, before a third panel copies it again — otherwise the next tweak to the caret glyph or label column width lands on one panel and silently misses the other.

internal/setting/settings.go
```diff
@@ -583,7 +583,7 @@ type OperationMode int
 const (
 	ModeNormal            OperationMode = iota
 	ModeAutoAccept                      // auto-approve edits/writes
-	ModeBypassPermissions               // allow all except deny rules and the root/home-removal circuit breaker
+	ModeBypassPermissions               // "YOLO mode": allow all except deny rules and the root/home-removal circuit breaker
```
hchenxa, line 586: While rewording this const comment (and the docs — permission-model.md now says "It is opt-out: unset means YOLO mode is in the cycle"), the adjacent comment above `cycleModes` (lines 592-594) still claims "BypassPermissions is only reachable when explicitly enabled" — the opposite of the opt-out default implemented in `Settings.AllowBypass()` (nil → true, so `NextWithBypass` includes YOLO in the default shift+tab cycle).

Suggest updating it to match, e.g. "BypassPermissions is in the cycle unless `allowBypass: false` locks it out".

hchenxa (commented):
Review (medium effort): 5 findings. Two correctness issues in the new `AllowBypassSavedMsg` handler — both stem from acting on the panel's user-level write intent instead of the effective merged value after reload — plus a status-line regression for non-OpenAI providers, one duplication, and one stale comment. Details inline.

hchenxa (changes requested):

hchenxa (approved):

PR #473, opened 2026-08-14
fix(deepseek): correct V4 pricing and default to the API's own thinking effort

yanmxa:
Two DeepSeek defaults were wrong against the published API.

- **Pricing.** V4 Pro was priced at 4x the real rate (1.74/3.48 vs 0.435/0.87 per million tokens), and both models billed cache hits near the cache-miss rate (Pro 0.145 vs 0.003625, Flash 0.028 vs 0.0028). `/cost` overstated DeepSeek spend accordingly. Rates now match [the pricing page](https://api-docs.deepseek.com/quick_start/pricing).
- **Thinking default.** DeepSeek runs thinking by default at effort `high`; san returned `"off"`, which actively sends `thinking:{type:"disabled"}` on every turn. A fresh session now gets what the API advertises.

Cost tests are now a table pinning one rate per case, including the cache-hit rate that regressed.

Not handled here: DeepSeek moves to peak/off-peak billing on 2026-08-16 (off-peak at half rate). The catalog is a single static rate per model, so off-peak turns will read 2x high until that's modelled — worth a follow-up.

### DIFF
--- a/internal/llm/deepseek/catalog.go
+++ b/internal/llm/deepseek/catalog.go
@@ -14,6 +14,10 @@ type modelCatalogEntry struct {
 	pricing pricing
 }
 
+// Prices are USD per million tokens, from
+// https://api-docs.deepseek.com/quick_start/pricing — inputPerMTokens is the
+// cache-miss rate and cacheReadPerMTokens the cache-hit rate. DeepSeek does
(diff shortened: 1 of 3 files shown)


hchenxa (approved):
/lgtm

hchenxa (approved):

hchenxa (approved):
/lgtm

PR #482, opened 2026-08-19
ci: scan for reachable vulnerabilities and track Go dependencies

yanmxa:
## What's exposed today

Dependabot covers `github-actions` only, and nothing scans the module graph. `govulncheck` on `main` reports **16 vulnerabilities this code can actually reach**:

```
Your code is affected by 16 vulnerabilities from 2 modules and the Go standard library.
```

Two of them are modules:

| module | | advisory |
| --- | --- | --- |
| `golang.org/x/text` | v0.37.0 → **v0.39.0** | GO-2026-5970 — infinite loop on invalid input |
| `github.com/yuin/goldmark` | v1.7.13 → **v1.7.17** | GO-2026-5320 — XSS |

The other fourteen are the CI toolchain pin. `1.25.8` is five patch releases behind, and I verified that **1.25.13 carries every one of those fixes**:

```
$ GOTOOLCHAIN=go1.25.13 govulncheck ./...     # before the module bumps
Your code is affected by 2 vulnerabilities from 2 modules.   # ← only the two above remain
```

The pin moves **within the same minor line**, so nothing about the language version, `go.mod`, or the supported build environment changes. `release.yml` moves with it — release builds are what ships to users.

## After the three bumps

```
$ GOTOOLCHAIN=go1.25.13 make vulncheck
No vulnerabilities found.
Your code is affected by 0 vulnerabilities.
```

So the new job **gates** rather than warns.

## The gate

`govulncheck` reports only vulnerabilities the code can *reach* — an advisory in a package nothing calls stays quiet. That reachability filter is what makes it worth blocking on: a finding is a real exposure, not a line in a triage backlog. (It stayed silent here on 1 vulnerability in imported packages and 2 in required modules, none of them reachable.)

- **Its own job**, not a step in `test` — a freshly published advisory is not a test failure and should not read as one on the PR page.
- **Weekly schedule** as well as per-PR, because an advisory published against a dependency already in `go.mod` would otherwise wait for someone to open a PR.
- **`make vulncheck`** runs the same command locally, with a pinned `govulncheck` version, matching the existing `check-format-tools` install pattern.

## Dependabot

Adds the `gomod` ecosystem, grouped by minor+patch so a week of bumps arrives as one PR rather than a dozen — the point is that a security fix lands quickly, not that every module gets its own review. Same weekly cadence, labels and limit as the existing github-actions entry.

Dependabot does **not** track the toolchain pin. `docs/operations/development.md` now says which file to edit when the scan points at the standard library, and notes that the stdlib half of a local run reflects the developer's own Go rather than the CI pin.

## Note on staying on 1.25

I kept this to a patch bump inside the pinned line, since moving to 1.26 is a version-policy call rather than a security fix. Worth knowing: Go 1.26.6 is the current stable, and the full test suite already passes under Go 1.26.2 locally, so that move looks low-risk whenever you want it.

## Verification

`GOTOOLCHAIN=go1.25.13 make vulncheck` → `No vulnerabilities found.` · `make lint` and `go test ./...` pass.


hchenxa:
@yanmxa please check the conflicts

yanmxa:
> @yanmxa please check the conflicts

Done

hchenxa:
@yanmxa , 2 files conflict, please take a look

PR #495, opened 2026-09-01
refactor(core): reach the model through one stream, not three

yanmxa:
San reaches a model through four layers today:

```
core.Agent → core.LLM → llm.Client → vendorProvider → ai.Client → ai.Driver
                ↑            ↑                            ↑
           core.Chunk   llm.StreamChunk               ai.Event
```

The bottom of that is already the SDK — `vendor.go`'s own header says so — which makes everything above it a second abstraction of the same thing.

## What the middle was doing

The same stream is reshaped twice on the way back. `vendorProvider.Stream` runs a goroutine turning `ai.Event` into `llm.StreamChunk`; `llm.Client.Infer` runs another turning `llm.StreamChunk` into `core.Chunk`. **Two goroutines and two channel hops on top of an iterator that needs neither.**

And `core.Chunk` cannot carry what the wire sends: two named string fields where a block has a type and an index, so an image or a tool call has nowhere to go and arrives only at `Done`.

`llm.Provider` also had exactly one implementation.

## After

`core.agent` ranges an `*ai.Client` directly.

Deleted: `core.LLM`, `core.InferRequest`, `core.Chunk`, `llm.StreamChunk`, `llm.Provider.Stream`, `llm.Client.Infer`, `llm.Client.Stream`.

`Provider` keeps the part San actually owns — reaching an endpoint — as `Client(model, headers) (*ai.Client, error)`.

The stall watchdog becomes `context.WithCancelCause` plus a timer instead of a `select` over three cases: an iterator cannot be selected on, and the cause is what carries the reason through a cancellation that would otherwise destroy it.

## What the seam still has to carry

Removing `Provider.Stream` removes the only place three things were happening. Each is now supplied where it belongs, and each has a test that fails if it stops being sent — the reason none of them failed one before is that the doubles inject core's own values, never a provider's.

**Classification.** `Classify`/`classifyStream` had exactly two call sites, both on the deleted path, so a provider failure reached the turn loop untagged. `*ai.Error` satisfies neither `core.RetryableError` (`RetryAfter` is a field, not a method) nor `core.ContextExceededError`, so `errors.As` never matched: a 429 or a 529 was fatal instead of retried, and `isPromptTooLong` was permanently false, which made the compact-and-retry path dead code. The classifier moves to `internal/core/classify.go` and runs where the failure leaves the stream — there is no package boundary in between for it to arrive untagged across.

**Settings.** `requestOptions` went with the provider and nothing replaced it: `ai.WithEffort` had zero call sites. The `/model` rung picker, its shortcut, its per-model persistence and `SetThinkingEffort` were all decoration, and the output cap fell back to the model's own `MaxOutput`. `Config.CallOptions` supplies them, asked for fresh on each inference so a mid-session change lands on the next call.

**Headers.** Every `Provider.Client` caller passed `nil`, leaving `turnHeaders` assigned and never read. Copilot rejects image content outright unless the request opts into vision, so sending a picture there could not work. `Config.Client` is therefore a function of the conversation, which is the honest shape — which client answers is a per-turn question — and `TurnHeaders` is the optional provider extension that answers it, beside `SupportsImages` and `CachesToolsAndSystemPrompt`.

`InputLimit` moves the same way, from a method on `core.LLM` to a `func() int` on `Config`. It never belonged to an inference abstraction — the figure San uses is the model's window *unless a setting overrides it*, and that setting is the application's.

## Also in here

Scrollback fixes that are not part of the refactor, kept here because they landed on this branch. All three are one story: `insertAbove` preserves the managed frame around a `Println`, so what that frame contains at that moment ends up in native terminal history.

- **Docked prompts stay out of terminal history.** A scrollback handoff while an approval, question or secret panel is up writes that transient panel into history. The queue now waits for the panel to close, a permission prompt waits for the queue to drain, and a docked modal renders on the alternate screen so a tall frame cannot scroll through the primary buffer on its own.
- **A held print is resumed by any panel closing, not just the three prompts.** The hold is on all sixteen overlays but the resume was wired to the three prompt replies, so a print queued while a slash-command picker was up was never restarted. The queue is single-flight, so every later block sat behind the stalled head — and `renderAndCommit` advances `CommittedCount` before the print runs, so the live tail had already stopped drawing what the queue was still holding. Open `/config` mid-answer and that answer was on neither.
- **The streaming tail is no longer welded into history.** An answer arrived twice: once as the plain, unrendered tail with the spinner still on it, and again underneath as the rendered block. San never printed the first one — committing empties the live tail before the print runs, and Bubble Tea's inline renderer redraws only the frame's new extent, leaving the rows the taller frame occupied on screen. The handoff copy that fixes it was here before, in e02ed422; 74ac1647 dropped it on the reading that the print FIFO and the renderer's flush barrier had made it redundant. They had not: the barrier fixes *when* `insertAbove` sees the frame, this fixes *what* the frame is.
- **The bidirectional hook tests stop racing a wall clock.** The 500 ms stdin grace period is a field on `Engine` so a test can set it instead of hoping the machine is idle.

## Verifying

The scrollback fixes were checked against a scripted endpoint in tmux: three turns at 24 and 50 terminal rows, three duplicated answers before and none after, each answer rendered exactly once.

All 66 packages pass, plus `go vet`, `golangci-lint` (0 issues) and `layercheck`. Every test double that faked `core.LLM` or `Provider.Stream` now fakes `ai.Driver`: a better seam, since it is the protocol, and five real implementations sit behind it. `legacyStream` — the chunk stream kept for the vendor tests — reaches the endpoint the way production does, headers and call options included, so a setting that stops being sent fails a test rather than going quiet on the wire.

Requires sdk-go v0.2.0 (first commit), whose only breaking change for San is one three-line `ai.Tool` literal.

Net **+1741/−923** across 57 files.


yanmxa:
## The `test` failure is pre-existing, not from this change

`internal/hook`'s `TestHooks_BidirectionalPrompt_SingleRound` fails on the Linux runner. I checked it against `main` rather than assuming:

| | full suite under `-race` | that test alone |
| --- | --- | --- |
| `upstream/main`, clean worktree | **FAIL** — `SingleRound` 0.76s, `UserDeclines` 0.77s | 10 runs, all pass |
| this branch | **FAIL** — `SingleRound` 1.00s | 5 runs, all pass |

Both fail the same way on an untouched `main`, so it is not this PR. `test-windows` passes on this branch, which fits: the failure is a timing flake that only shows up under the contention of the full race-enabled run on the slower runner, and every failure lands under one second.

`internal/hook` is not in this diff. The one hook-adjacent file that is — `internal/app/hooks_test.go` — is there only because `switchProviderStub` had to become an `ai.Driver` like every other test double.

Happy to open a separate issue for the flake if that is useful; it looked out of scope for a refactor that is meant to change shape and nothing else.

PR #498, opened 2026-09-03
refactor(core)!: the conversation is ai.Message

yanmxa:
The first half of retiring `internal/core`: San's conversation type becomes the
SDK's, so `pkg/agent` can hold it. Nine commits, each green on its own.

## What was in the way

Not the loop — the **types**. `core.Message` held eleven parallel fields, and
four other types differed from `pkg/ai`'s by one field each. Every one of those
differences turned out to be a modelling error, held in place by nothing but
habit:

| | what was wrong | where it went |
|---|---|---|
| `ToolCall` | `ThoughtSignature` vs `Signature` | alias — **135 call sites unchanged** |
| `Image` | `Path` is provenance, not a picture; `Size` was dead | `Attachment{ai.Image; Path}` |
| `Role` | a notice is not a role — the view has 3 producers, the conversation 2 | `ChatRole` + `ai.Role` |
| `ToolResult` | `HookResponse` dead; `Details` is the interface's | `ToolDetails` on `ChatMessage` |
| `Message` | 11 parallel fields guessing an order the data already had | `= ai.Message` |

## The one that was hiding something

`Details` **couldn't** move to `ChatMessage`, because persistence read it off
the conversation:

```
ChatMessage → MessagesFromChat → core.Message → MessageToBlocks → node
```

The transcript — a record of *what the interface showed* — was routed through
the type that exists to be **sent to a provider**. So anything the transcript
needed had to be bolted onto the conversation to survive the trip;
`DisplayContent` had already been bolted on the same way.

The transcript now converts against `ChatMessage` directly and
`Snapshot.Messages` is `[]core.ChatMessage`. `chat_convert.go` is deleted whole;
`model_session.go` saves and restores with **no conversion at all**.

## Deleted because the gap closed

- `ToAIMessages` — had become the identity function
- `llm.toProviderMessages` — copied "only the fields a provider needs"; there
  are none left to leave behind
- `replayableThinking` — the per-protocol replay table, now `ai.Model`'s
  (sdk-go v0.4.1, which this branch also caused)
- the `[Image #N]` token dance on the provider path — `ai.Content` interleaves

## Two silent bugs the shape fixed

- **`ChatMessage` had no `Reasoning`**, so a conversation that went through the
  view and back lost its opaque state and a reasoning model started over.
- **`Signal` rode on `Message`** because the inbox was a `chan Message`. It is
  `Inbound` now, and goes away entirely with the loop.

---

suite passes. `go.mod` is on the released **v0.4.1** — no `replace`.


yanmxa:
Superseded by #499 — closing.

#499 was branched on top of this one, and san-ci squash-merges: merging it collapsed both branches into `6918070f`, so everything here is already on main.

```
main:internal/core/message.go:90   type Message = ai.Message      ← this PR
main:internal/core/tool.go:14      type Tool = sdkagent.Tool      ← #499
```

Merging this branch now would *revert* #499 — `git diff main origin/sdk-migration-messages` is 53 files, +238/−284, all of it putting `Tool` back to an interface, `ReasoningItem` back to San's own struct, and `ToolResult.Content` back to a string.

The rest of the stack (#500 → #501 → #502) is unaffected: each of those trees is a superset of main and reverts nothing.

PR #500, opened 2026-09-04
refactor(core)!: the loop is the SDK's

yanmxa:
One commit on top of main, no stack to read first. (#499's squash carried
#498's commits with it, so #498 is closed as superseded.)

`ThinkAct` no longer *runs* an exchange — it drives one and translates what
comes back. 18 files, **−243 lines net**, and not one of the ten packages that
read the event stream changed *how* it reads one. (Four files in them do
change: three drop a `Config` field that had gone dead, one is the `All()`
defect below.)

## What moved

The inference and its retries, the tool batch and its parallelism, the
compaction check: all `sdkagent.Agent` now. What is left is the part the SDK
deliberately does not have — a **mailbox**. `Run` blocks on an inbox, drains
what accumulated, hands the batch to one exchange. How messages are batched is
the application's to decide.

| `core.Config` | became |
|---|---|
| `Client` | `PreInfer` sets `Inference.Client` — still per turn |
| `CallOptions` | `PreInfer` appends to `Inference.Options` |
| `InputLimit` + `CompactFunc` | `PreStep`, and `OnInferError` for the too-long prompt |
| `MaxTurnRetries` | `WithRetry`, **+1** — San counts retries, the SDK counts attempts |
| `MaxOutputRecovery` | `WithContinuation`, with San's own prompt |
| stream timeouts | `WithStreamTimeout` |

## Two figures stopped needing to be kept

`lastTotalInputTokens` is gone, and with it the trap it existed to dodge: it had
to be cleared by hand after a compaction or the stale number would still read
"full" against the conversation that had just replaced it — and compact again on
the very next step, forever. `PreStepContext.Tokens` is measured fresh at each
boundary.

Batch parallelism went the same way. The rule used to be checked when a batch
arrived — all read-only, or all agent-spawning, or run them one at a time; each
tool now carries the answer into the loop as `sdkagent.Sequential`, so a batch
is parallel exactly when every member says it may be and the loop never asks
about the batch as a whole. *Which* tools those are is still a list in `core`,
because nothing in a tool's schema distinguishes reading a file from writing
one.

## Two defects this surfaced

Both were invisible while the loop fetched tools one at a time by name:

1. **The call's ID stopped reaching tools.** The loop used to put it in the
   context before dispatching. Restored by a wrapper.
2. **The decorating registries wrapped on `Get` and not on `All`.** Permission,
   PreToolUse and subagent-activity all did this — asking for the whole set
   returned *undecorated* tools, so a batch fetched that way would run
   **ungated**. Nothing asked that way before; the first thing that did found
   it.

## The naming pass folded in

The sweep that used to sit in #501 belongs here, in the commit that introduced
what it renames — a name that lives for exactly one commit is worse than the
wrong name.

- `agent_impl.go` was named for being an implementation, and held **three**
  subjects once the loop left. **`run.go`** is the mailbox — an inbox, an
  outbox, and the loop that waits on one and reports to the other.
  **`exchange.go`** is one exchange: the toolset it offers, the hooks it
  installs, driving the SDK's loop through it, and turning every event into
  San's. **`compact.go`** is shortening a conversation that has outgrown its
  window — which belongs to *neither*, since `run.go` calls it for `/compact`
  and `exchange.go` calls it from two hooks. Leaving it in either would have
  repeated `agent_impl.go`'s mistake one file down.
- `unreachable` → **`unconfigured`**: its own model ID, driver name and error
  text all said "unconfigured" — three against one.
- `identified` → **`withCallID`**: the tool is not what is identified; the call
  is.
- `sequenced()` → **`offered()`**: it read as though every tool came back
  sequential. It is what the model may call this exchange, with the ones that
  must not run beside others marked.
- `retryBaseDelay` → **`backoffBase`**: it is the base of a backoff.

Renaming `retry.go` itself stays in #501, where the retry vocabulary leaves —
renaming it here would be renaming a file that still had retrying in it.

**Three things nobody read**, all dead because of this change: `Config.CWD`
(four setters, zero readers — a tool has resolved its own working directory
through a closure for a long time), `Config.MaxOutputRecovery` (still copied
onto the agent, unread since the budget became `WithContinuation`'s), and
`BuildConversationText` / `conversationTextLen` (no caller outside their own
tests — the old loop sized a prompt with them, and sizing a prompt is the
SDK's now).

---

suite passes.


yanmxa:
**Fix pushed** — found reviewing the stack, so it lands in the commit that introduced it.

`stopReasonOf` sent everything it did not name to `StopError`, and the SDK names two reasons San does not: `StopRefusal` (the model declined, or a content filter fired) and `StopSequence` (generation hit a caller's stop sequence). Both end a turn.

`run.go` reads `StopError` to mean the agent died and skips `TurnEvent` for it, so a refused turn would have produced **no turn boundary at all** — the queue would not advance, `OnTurnEnd` would not fire, and autopilot would never see the turn end. The mapping this branch replaced got it right; the new one lost the rule.

Now only `sdkagent.StopError` is `StopError`, and the default is `StopEndTurn` — the safer default, since a future SDK reason landing there ends a turn instead of hanging one. `TestOnlyAFailedInferenceIsAnError` covers all eight.

PR #501, opened 2026-09-04
refactor(core): San kept a second answer to a question the SDK answers

yanmxa:
**Stacked on #500** — one commit, reviewed after it.

`classify.go` said why it existed at the top: translate the SDK's reading of a
provider failure onto two interfaces of San's own, *"so the loop carries no
provider vocabulary"*. **The loop is the SDK's now** (#500) and reads
`ai.IsRetryable` directly — so the translation had nobody left to translate
for, and `RetryableError` / `ContextExceededError` were a second vocabulary for
a question already answered.

The two places outside the loop that asked it now ask the SDK. `llm`'s one-shot
`Complete` keeps classifying on the way out, for the same reason it always did
— an error leaving that package is always classified, so no caller gets a
different answer by reaching for a different helper — except that what it hands
back is now an `*ai.Error` rather than a San wrapper around one.

## And the file named after them

`retry.go` → `backoff.go`. A file claiming to be about retrying was fair while
it held the retry vocabulary; what is left in it is a backoff with a cancel,
for the two retry loops that are **not** the agent's — `llm`'s one-shot
`Complete` and autopilot's steering call. Renaming it a commit earlier would
have been renaming a file that still had retrying in it, which is why the rest
of the naming pass lives in #500 and this half lives here.

**259 lines of vocabulary out**, and 371 out against 59 in across the whole
commit once `retry.go` goes with them.

## Verification

suite passes, and the commit builds and tests on its own.


yanmxa:
Superseded by #502 — closing.

#502 was branched on top of this one, and san-ci squash-merges: merging it collapsed both branches into `a91172c6`, so everything here is already on main. Applying this PR's whole diff to main changes **0 files**.

```
main:internal/core/agent.go   OnCompactEnd  ← this PR (the compaction span fix)
main:internal/core/classify.go              ← deleted by this PR
```

Same shape as #498 → #499. The follow-up that adapts the application layer directly onto the SDK's events is next, on top of main.

PR #512, opened 2026-09-10
fix: serialize agent generations and stop losing hook context, forked todos, and task metadata

yanmxa:
## What & why

Four defects in the agent / session / task runtime, each with a user-visible
symptom, plus the boundary changes that make them fixable without special
cases. Two commits: the first carries the changes, the second is a cleanup
pass over the first (dead code, no-op injection, test-only wrappers).

### Fix 1 — a stopped agent could still be running when its replacement started

| | |
| --- | --- |
| **Symptom** | After `/fork`, the parent transcript keeps growing with the fork's messages, inferences, permissions and hooks. Resuming the parent (`san -r <id>`, which the fork banner tells the user to do) replays post-fork history. |
| **Cause** | `Session.Stop` cleared the agent handle and returned immediately; the `Run` goroutine and its recorder — bound to the parent session ID — kept going while `Start` built the new agent. The old code carried a comment admitting the stop/restart race. |
| **Change** | A run generation (`sessionRun`: agent, cancel, done) is owned until `Run` returns. `Stop` cancels and waits for that exact generation (2 s bound; `StopContext` for other deadlines). `Start` refuses while a generation is alive. `Active()` flips only after the goroutine exits. |
| **Test** | `TestSessionStopWaitsForExactRunBeforeRestart`, `TestSessionUnexpectedExitClearsLifecycle` (`internal/agent/session_test.go`) |

### Fix 2 — async hook findings never reached a running agent

| | |
| --- | --- |
| **Symptom** | An `asyncRewake` hook reports context lines (policy finding, tool detail) plus a continuation prompt. The lines show up in scrollback, but the agent only receives the prompt and continues without the findings. |
| **Cause** | `injectAsyncHookContinuation` appended each context line to the conversation for display and submitted only `ContinuationPrompt`. |
| **Change** | Context and continuation are joined into one user turn that is both displayed and delivered. |
| **Test** | `TestAsyncHookContinuationKeepsContextInDeliveredTurn` (`internal/app/model_turn_queue_identity_test.go`) |

Same path, related: `Session.Send` used to return nothing and silently drop
the message when no agent was active. It now returns `ErrSessionInactive` /
`ErrSessionStopped`, and the app surfaces a delivery failure as a notice.

### Fix 3 — `/fork` lost the todo list

| | |
| --- | --- |
| **Symptom** | Todo items (and their in-progress state) disappear after `/fork`. |
| **Cause** | `forkSession` cleared the tracker storage dir, then `initTaskStorage` pointed it at the fork's directory — which is empty — and `SetStorageDir` loads from disk, replacing the in-memory items. Nothing re-imported the forked snapshot's tasks. |
| **Change** | After the storage switch, `Tracker.Import(forked.Tasks)`; the same sequence `loadSessionByID` already uses. |

### Fix 4 — task metadata was dropped on persist

| | |
| --- | --- |
| **Symptom** | After resume, a todo item no longer knows which background worker is advancing it (`task_id` gone), so orphan demotion and the task panel lose the link. |
| **Cause** | `transcript.TrackerItemView` had no `Metadata` field; the hand-written converters silently dropped it. |
| **Change** | `TrackerItemView` mirrors `todo.Item` field for field, so the converters are plain struct conversions — a future field mismatch fails to compile instead of dropping data. JSON tags match `todo.Item`, so the on-disk `PatchTasks` payload is unchanged. |
| **Test** | `TestTrackerItemViewsRoundTripPreservesMetadata` (`internal/session/task_views_test.go`), `TestProjectTasksAndWorktreePatches` (metadata assertion) |

### Supporting change — one message ID from projection to inbox

Fix 1 restarts the agent from a preloaded chain and must drop the message the
caller is about to re-send. That dedupe used to compare message *text*
(`last.Text() == pendingSend`), which is wrong whenever a reminder was
attached or two identical messages were sent. Now:

- `conv.Append` returns the stamped `ChatMessage` (`TestAppendReturnsStoredMessageIdentity`);
- every submit site builds its `core.Message` from that row, so the TUI
  projection, the agent inbox and the transcript share one ID
  (`TestSessionSendPreservesMessageIdentity`);
- `ensureAgentSession(pendingMessageID)` dedupes by ID.

`SubmitToAgent` / `Session.Send` take a `core.Message` instead of
`(content, images)`.

### Hardening

- **Hooks.** Command hooks opt into the `PromptRequest`/`PromptResponse` line
  protocol with `"interactive": true`. Non-interactive hooks get one JSON
  value and EOF on stdin, replacing the 500 ms idle timer that guessed which
  mode a script wanted (`TestHooks_NonInteractiveCommandClosesStdinWithPromptCallback`).
  Both modes run detached from the TUI's controlling terminal
  (`proc.DetachSession`), so a script that reads `/dev/tty` fails fast instead
  of hanging the UI. `setting.Data.Clone` is now a value copy — the
  field-by-field copy would have dropped `Interactive` on the way to the
  engine — and the plugin `hooks.json` loader maps the key too.
- **Task output.** The output directory lives on each `task.Manager` instead
  of a package global (`TestManagersKeepOutputDirectoriesIndependent`); tasks
  create their log file outside the registry lock.
- **Todo storage.** `SetStorageDir` prepares the directory before adopting it,
  so a failed reconfiguration leaves the store on its previous directory with
  its items intact (`TestSetStorageDirRollsBackAfterConfigurationFailure`).
- **Layering.** `transcript` no longer imports `todo`. `layercheck` applies
  the layer rules to test imports and rejects unmapped packages within its
  `go list` scope. `make build` no longer rewrites sources; `make release`
  runs `format-check`.

### Breaking change

Command hooks that implement the prompt protocol must set
`"interactive": true`; without it the engine closes stdin after the initial
input and the script sees EOF where it expected a `PromptResponse`.

### Review map

| Area | Files | What to look at |
| --- | --- | --- |
| Fix 1 | `internal/agent/session.go` | `sessionRun`, `execute`, `Stop`/`StopContext`, `Send` |
| Fix 2 | `internal/app/model_turn_queue.go`, `internal/app/update.go`, `internal/app/agent.go` | `injectAsyncHookContinuation`; `agentSendFailedMsg` |
| Fix 3 | `internal/app/model_session.go` | `forkSession` |
| Fix 4 | `internal/session/task_views.go`, `internal/session/transcript/{types,projector,store,view}.go` | struct conversion; JSON tags on `TrackerItemView` |
| Message ID | `internal/app/conv/conversation.go`, `internal/app/update_submit.go`, `internal/app/input/slash_command.go` | `Append` return; `SubmitToAgent(core.Message)`; `ensureAgentSession(pendingMessageID)` |
| Hooks | `internal/hook/{executor,executors_command,engine}.go`, `internal/setting/settings.go`, `internal/plugin/resolver.go`, `internal/proc/` | `Interactive` gate; `buildShellCommand`; `Clone` |
| Task output | `internal/task/{manager,output_store,agent_task,bash_task}.go`, `internal/subagent/executor.go` | `outputPath`; constructors take the path |
| Todo storage | `internal/todo/store.go` | `SetStorageDir` ordering |
| Tooling / docs | `Makefile`, `tools/layercheck/main.go`, `docs/**` | |

## Type of change

- [x] Bug fix
- [ ] New feature
- [x] Refactor / cleanup
- [x] Docs
- [ ] Other

## Testing

```bash
make ci
GOOS=windows GOARCH=amd64 go build ./cmd/san/
git diff --check
```

`make ci` passed locally: build, vet, golangci-lint, format check, layer
check, race-enabled unit tests, and the integration suite. The tests named
above are new in this PR.

## Checklist

- [x] Tests pass locally


yanmxa:
Superseded by #518, which carries the same latest rebased and reviewed commits on a fresh head branch. Continuing review there.

PR #515, opened 2026-09-12
fix(llm): open the sign-in page and report the current Codex client version

mannsion:
## Summary

Two fixes to the ChatGPT subscription endpoint:

- **Sign-in never opened the browser.** When interactive sign-in moved to the SDK (`auth.Login`), San's custom `Interaction` surfaced the URL in the TUI but dropped the `proc.OpenURL` call the old in-tree flows made. Restore it so the browser opens; the TUI prompt remains for headless/SSH cases.
- **Newer models were invisible.** The ChatGPT model catalog gates its lineup behind the `client_version` the caller reports. San reported Codex `0.144.0`; the current Codex CLI is `0.154.0`. Bump the constant so the catalog serves models gated on newer clients.

Also gofmts `internal/app/conv/context_usage.go` (indentation only).

## Testing

- `make ci` (vet, format-check, golangci-lint, layercheck, race tests) passes.
- `go test ./internal/llm/ ./internal/app/input/` passes.
- Verified live against a ChatGPT subscription: `client_version=0.144.0` lists 7 models; `0.154.0` lists 8, including `gpt-6-astra` (absent before, present after).


yanmxa (approved):
Thank you so much for following up on this — both fixes are exactly right, and the live 7 → 8 model check is much appreciated.

- The `proc.OpenURL` call is a faithful restore of what the old in-tree Copilot/OpenAI flows did before #485 moved sign-in into the SDK; best-effort with the TUI prompt as fallback is the right shape.
- `0.154.0` matches the current Codex stable line (0.155 is still alpha).

One note on the third commit: your gofmt is Go 1.27's, which indents composite literals in a multi-value `return` differently from 1.26 — CI here was still pinned to 1.26.8 and would have flagged the reformatted file. That's on us, not you: #516 bumps CI to 1.27 and carries the same reformat, so once it lands this PR will go green on a rebase. Nothing you need to change.

Approving now; will merge as soon as CI is through.

PR #519, opened 2026-09-13
fix(session): persist every result of a parallel tool call

yanmxa:
## What & why

A turn that answers N parallel tool calls is one agent message carrying N `tool_result` blocks (`ai.ToolResultsMessage(results...)`). `core.ChatOf` projects a message onto a `ChatMessage`, which has room for one `ToolResult`, so the recorder wrote only the **last** result into the transcript.

Reproduced in tmux against a stub provider: the model issues two `TaskCreate` calls in one response; the transcript ends up with

```
tool_use   TaskCreate call_0
tool_use   TaskCreate call_1
tool_result call_1            ← call_0's result is gone
```

On `san -r` the replayed chain then carries a `tool_use` with no answer, the SDK's `ai.Repair` pass drops it together with its result, and the provider receives `a(tc:1) t` where the live session had `a(tc:2) t t`. The model silently loses part of what it did; with a stricter provider the orphaned block would be a request error instead. Subagent transcripts (`SaveSubagentConversation` → `chatOf`) lost results the same way.

### Change

- `core.ChatRowsOf(m)` — the projection that keeps every result: one row per `tool_result`, all sharing the turn's ID. `ChatOf` is unchanged (and now documents that it keeps only the last result).
- Recorder `onAppend` writes the blocks of all rows into the turn's **single** `message.appended` record — the on-disk shape is still one record per turn, and matches the wire format.
- `messagesFromNodes` splits a node with several `tool_result` blocks into one row per result (with `ToolName` backfilled per row); `messagesToNodes` folds sibling rows (same ID, tool results) back into one node — the inverse, so a re-save of a loaded session writes nothing twice.
- `Store.chatOf` (subagent save) uses the rows.

### Test

`TestParallelToolResultsSurviveRecordAndLoad` (`internal/session/parallel_tool_results_test.go`): record a two-result turn → load gives two rows with the right IDs, text and tool names → folding back yields one node with two blocks → re-save + reload is stable. Fails on `main` with `tool result rows = [c2=two/Bash], want [c1=one/Read c2=two/Bash]`.

Verified end to end in tmux: after this change the resumed chain sent to the provider is `s u a(tc:2) t t a(tc:1) t a u`.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor / cleanup
- [ ] Docs
- [ ] Other:

## Testing

```bash
go test ./internal/session/... ./internal/core/... ./internal/app/... ./internal/inspector/...
make lint-go lint-layers
git diff --check
```

(`make format-check` disagrees with `internal/app/conv/context_usage.go` under local Go 1.26 gofmt — untouched here; CI's Go 1.27 formats it the way it is.)

## Checklist

- [x] Tests pass locally


hchenxa:
test comments

PR #521, opened 2026-09-13
chore: bump version to 1.22.8

yanmxa:
Release v1.22.8 with an attributed changelog and source version bump. Includes the current Bash description rendering changes: use dimmed parentheses and preserve the closing parenthesis when truncating. Validation: go test ./internal/app/conv passed. Release metadata will be added before marking ready.


yanmxa:
Release canceled at the author's request. The Bash description change will be submitted separately on a non-release branch; no version bump or tag is intended.

